PR #2, opened 2021-03-24
chore: add python repo templates

busunkim96:
Add some python repo templates via synthtool to enable the usual release flow (release-please and Kokoro release builds). 

Needs corresponding kokoro config change internally.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,7 @@
-Before you open a pull request, note that this repository is forked from [here](https://github.com/docascode/sphinx-docfx-yaml/).
-Unless the issue you're trying to solve is unique to this specific repository, 
-please file an issue and/or send changes upstream to the original as well.
```
dandhlee, line 3: Could you preserve these 3 lines?

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
+# How to Contribute
```
dandhlee, line 1: we already have CONTRIBUTING.md in the top directory, should we move that under `.github` directory?

setup.py
```diff
@@ -30,7 +30,7 @@
     version='1.2.76',
     author='Eric Holscher',
     author_email='eric@ericholscher.com',
-    url='https://github.com/ericholscher/sphinx-docfx-yaml',
+    url='https://github.com/googleapis/sphinx-docfx-yaml',
```
dandhlee, line 33: Could you update the file with the following?
```
import codecs
import setuptools

name = 'gcp-sphinx-docfx-yaml'
description = 'Sphinx Python Domain to DocFX YAML Generator'
version = '0.1.0'
dependencies = [
    'PyYAML',
    'wheel>=0.24.0',
    'sphinx',
    'unidecode'
]

packages = setuptools.find_packages('.', exclude=['tests'])

extra_setup = dict(
setup_requires=['pytest-runner'],
tests_require=['pytest', 'mock'],
)

setuptools.setup(
    name=name,
    version=version,
    description=description,
    author='Google LLC',
    author_email='dandhlee@google.com',
    license='Apache 2.0',
    url='https://github.com/googleapis/sphinx-docfx-yaml',
    package_dir={'': '.'},
    packages=packages,
    install_requires=dependencies,
    include_package_data=True,
    zip_safe=False,
    **extra_setup
)
```

dandhlee (changes requested):
Thank you so much for doing this! Just few minor changes would be great!

.github/PULL_REQUEST_TEMPLATE.md
busunkim96: Yep I'll add it to the `excludes` so it doesn't get overwritten

.github/CONTRIBUTING.md
busunkim96: Same comment as above, I'll just exclude this from the copy

dandhlee (approved):
Thank you!

PR #16, opened 2021-05-10
feat: clarify names in the left nav

### LINKED ISSUE #15: Make the table of content entries more short and concise
Currently it uses `uid` field instead of `name` from each module when inserting entries into the toc. The full `uid` usually is very long and thus makes it very hard for users to see what the entry in the left nav is. Making use of `name` or other similar fields or additional parsing will help.

For bigquery storage, it currently looks like
```
google-cloud-bigquery-storage
|- Overview
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
|- google.cloud.bigquery_storage_v...
```
before expanding.

dandhlee:

Currently it finds the parent node through `find_node_in_toc_tree` and found the parent by traversing through names and the parent. By using shorter names, it could no longer find the parent, so I had to make use of the additional `uidname` field. This is present in the entire toc structure; making use of only existing `uid` field fails to parse as some entry levels only contain the `name` and `items` field. 

The toc with uid gets used to further determine where entries go, and the one that gets stripped gets put in the `toc.yml` file with the content we want. Overall, we achieve shorter names, for example

```
google-cloud-spanner
|- Overview
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
|- google.cloud.spanner_v...
```
to
```
google-cloud-spanner
|- Overview
|- database_admin
|- spanner_admin_database_v1.types
|- instance_admin
|- spanner_admin_instance_v1.types
|- batch
|- client
|- database
|- instance
...
```

Fixes #15

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -564,12 +564,30 @@ def build_finished(app, exception):
     Output YAML on the file system.
     """
 
-    # removes uidname fields from the toc_yaml variable
-    def organize_toc_name(toc_yaml):
+    # Used to get rid of the uidname field for cleaner toc file.
+    def sanitize_uidname_field(toc_yaml):
         for module in toc_yaml:
             if 'items' in module:
-                organize_toc_name(module['items'])
+                sanitize_uidname_field(module['items'])
             module.pop('uidname')
+
+    # Parses the package name and returns package name and module name.
+    def find_package_name(package_name):
+        for name in package_name:
+            if name != "google" and name != "cloud":
+                return [name, package_name[-1]]
+
+    # Used to disambiguate names that have same entries.
+    def disambiguate_toc_name(toc_yaml):
+        names = {}
+        for module in toc_yaml:
+            names[module['name']] = 1 if module['name'] not in names else 2
+            if 'items' in module:
+                disambiguate_toc_name(module['items'])
+
+        for module in toc_yaml:
+            if names[module['name']] > 1:
+                module['name'] = ".".join(find_package_name(module['uidname'].split(".")))
 
     def find_node_in_toc_tree(toc_yaml, to_add_node):
         for module in toc_yaml:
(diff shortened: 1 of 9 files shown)


docfx_yaml/extension.py
```diff
@@ -756,6 +764,15 @@ def convert_module_to_package_if_needed(obj):
                     raise ValueError("Unable to dump object\n{0}".format(yaml_data)) from e
 
             file_name_set.add(filename)
+           
+            # Parse the name of the object.
+            # "Types" class will need additional parsing to de-duplicate their names and contain
+            # a portion of their parent name for better disambiguation.
+            node_name = obj['name']
+            if node_name == "types":
+                node_name = ".".join(obj.get('uid').split(".")[-2:])
```
tbpg, line 773: Are we sure the last two elements always contain what we want?

dandhlee: The name is represented in the last portion of the uid, and I just need one more tiebreaker, which the second last item will work out: we can have same module name but different package (`storage_v1.client`, `storage_v1beta2.client`) or different module and same package (`storage_v1.client`, `storage_v1.types`). Any more different than levels of two should appear under different directory structure, and that doesn't need a disambiguation.

However I should fix hardcoding types, and move this algorithm elsewhere so it detects duplicate entries beside `types`.

dandhlee:
`package.module` name can sometimes be long, but hard to shorten it concisely without losing the original context. Will work on that as an enhancement for later on.

tbpg (approved):
Thank you!

Issue #31, opened 2021-05-24
Allow generate-docs Kokoro job to run for multi-version releases on failures


dandhlee:
Currently, it is likely expected for the `generate-docs` Kokoro job to fail on multi-version releases when given the option for `GITHUB_TAG=all` is given. While the option is supported and will retrieve all the tags and run against all releases, there probably are releases where it doesn't have nox builds for `docs` or `docfx` jobs which would fail the run prematurely.

This issue should be closed if the Kokoro job doesn't fail on all the tags for every repo, or once it is updated to not fail on the tags (should skip on tags that we don't expect it to be upgraded on).

dandhlee:
Testing on Spanner, it fails on `1.17.1` tag since it's missing docfx. Will need to look into skipping the ones that fail.

tbpg:
We should be able to generate docs for every version, not just the ones where the DocFX or the docs job was already set up.

dandhlee:
That gives me an idea on how to work with this :O
Hypothesis: take the docfx setup, and apply only the relevant files from each version tag release. This should work for all versions that's simply missing docfx setup. This may not work for earlier versions that doesn't use Sphinx. Seems like the easiest first approach to take, will investigate further.

PR #32, opened 2021-05-24
ci: add YAML generation Kokoro job

dandhlee:

Adding Kokoro job config to allow generation of all the YAML throughout the Python Client Libraries.

Currently scoped to the repositories that have already submitted to the Google Cloud Storage bucket and runs against the latest release by default. Retrieves the repository information, clones into it, checks out the latest tag, performs the docs build and then moves onto the next repository once it's done uploading the built content.

I have implemented an early stage of supporting for multi-version release (processing every tag known for each repo) but it will probably fail, so I filed #31 to track for it.

- [X] Tests pass


dandhlee:
Updated to simply use `FORCE_GENERATE_ALL_TAGS` option to run against all tags or just on the latest, defaulted to `FORCE_GENERATE_ALL_TAGS="false"` to only force generate all YAMLs on the latest release.

.kokoro/generate-docs.sh
```diff
@@ -0,0 +1,106 @@
...
+    else
+      # Grab the latest released tag
+      GITHUB_TAGS=$(git describe --tags `git rev-list --tags --max-count=1`)
+    fi
+
+    # TODO: allow skipping failing docs builds and continue with the rest of the generation.
+    for tag in ${GITHUB_TAGS}; do
+      git checkout ${tag}
+
+      # Build HTML docs for googleapis.dev.
```
busunkim96, line 65: Is it necessary to also re-upload docs to googleapis.dev?

dandhlee: It isn't at the immediate scope of what I'm trying to do. I'll comment it out for now, but I don't think it'll hurt to support that option with like an environmental variable passed in case if anyone wants to regenerate all of the googleapis.dev content as well.

dandhlee:
Thank you! Fixed it to only process and upload YAMLs for Cloud RAD objects. Please take a look again.

busunkim96 (approved):
Thanks Dan!

.kokoro/generate-docs.sh
```diff
@@ -0,0 +1,107 @@
...
+python3 -m pip install --user --quiet --upgrade nox
+python3 -m pip install --user gcp-docuploader
+
+# Retrieve repositories to regenerate the YAML with.
+for bucket_item in $(gsutil ls gs://docs-staging-v2 | grep "docfx-python"); do
+
+  # Parse the library name.
+  library=$(echo ${bucket_item} | cut -d "-" -f 5)
+
+  # Only process unique entries.
```
tbpg, line 33: Can use `sort` to only process unique library entries. Something like this:

```
LIBRARIES=$(gsutil ls 'gs://docs-staging-v2/docfx-python*' | sort -u -t- -k5,5)
```

tbpg (approved):
A couple Bash suggestions, otherwise looks good.

.kokoro/generate-docs.sh
```diff
@@ -0,0 +1,106 @@
...
+export PYTHONUNBUFFERED=1
+
+export PATH="${HOME}/.local/bin:${PATH}"
+
+# Install dependencies.
+python3 -m pip install --user --quiet --upgrade nox
+python3 -m pip install --user gcp-docuploader
+
+# Retrieve repositories to regenerate the YAML with.
+for bucket_item in $(gsutil ls gs://docs-staging-v2 | grep "docfx-python"); do
```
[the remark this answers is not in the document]
dandhlee: done!

.kokoro/generate-docs.sh
dandhlee: This is much better, thank you!

PR #41, opened 2021-06-15
feat: Add support for Property and missing content

### LINKED ISSUE #40: Fix content missing compared to googleapis.dev
There's a lot of content that's missing when taking a look at googleapis.dev versus the YAML that gets generated with the sphinx plug-in tool. This seems like due to following:

- Sphinx upgrade from 1.x to 2.1.0+ separated the summary that's missing in newer versions
- The plugin has no support for the separated part of the summary

dandhlee:

While investigating for adding support for property, I managed to add a section for content discrepancy as both fo them seem to be missing with the newer Sphinx upgrade from 1.x to 2.1.0+

This PR achieves the following:
- Adds handler for property so that they properly show in the YAML hierarchy
- Adds handler for missing documentation snippets for each entries
  - for each documentation snippet, sanitizes it into useful subsections for display in HTML pages
  - currently has handlers for `:type`, `:rtype:`, `:returns:`, `:raises`, and `:param`

@busunkim96 I added handlers for subsections based on the 5 given above from `python-spanner`, for example:

```python
    def execute_sql(
        self,
        sql,
        params=None,
        param_types=None,
        query_mode=None,
        query_options=None,
        retry=google.api_core.gapic_v1.method.DEFAULT,
        timeout=google.api_core.gapic_v1.method.DEFAULT,
    ):  
        """Perform an ``ExecuteStreamingSql`` API request.

        :type sql: str
        :param sql: SQL query statement

        :type params: dict, {str -> column value}
        :param params: values for parameter replacement.  Keys must match
                       the names used in ``sql``.

        :type param_types:
            dict, {str -> :class:`~google.spanner.v1.types.TypeCode`}
        :param param_types: (Optional) explicit types for one or more param
                            values;  overrides default type detection on the
                            back-end.

        :type query_mode:
            :class:`~google.spanner.v1.types.ExecuteSqlRequest.QueryMode`
        :param query_mode: Mode governing return of results / query plan. See:
            `QueryMode <https://cloud.google.com/spanner/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest.QueryMode>`_.

        :type query_options:
            :class:`~google.cloud.spanner_v1.types.ExecuteSqlRequest.QueryOptions`
            or :class:`dict`
        :param query_options: (Optional) Options that are provided for query plan stability.

        :type retry: :class:`~google.api_core.retry.Retry`
        :param retry: (Optional) The retry settings for this request.

        :type timeout: float
        :param timeout: (Optional) The timeout for this request.

        :rtype: :class:`~google.cloud.spanner_v1.streamed.StreamedResultSet`
        :returns: a result set instance which can be used to consume rows.
        """
```
given with such documentation support. Should I expect the client libraries to be accustomed to this, or should I take a more general approach and try to guess based on given parameters or simply leave them blank if I'm not given these in the documentation?

Fixes #40 

- [x] Tests pass
- [ ] Send part of this change upstream for support in missing content / handler for property

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -36,6 +36,8 @@ from sphinx.util.console import darkgreen, bold
 from sphinx.util import ensuredir
 from sphinx.errors import ExtensionError
 from sphinx.util.nodes import make_refnode
+from sphinxcontrib.napoleon.docstring import GoogleDocstring
+from sphinxcontrib.napoleon import Config
 
 from .utils import transform_node, transform_string
 from .settings import API_ROOT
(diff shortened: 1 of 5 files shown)


docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
...
+    if lines != []:
         lines = _resolve_reference_in_module_summary(lines)
         summary = app.docfx_transform_string('\n'.join(_refact_example_in_module_summary(lines)))
+    
+        # Extract summary info into respective sections.
         if summary:
-            datam['summary'] = summary.strip(" \n\r\r")
+            top_summary = ""
+
+            # Initialize known types needing further process.
```
tbpg, line 400:
```suggestion
            # Initialize known types needing further processing.
```

docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
         'langs': ['python'],
     }
 
-    # Only add summary to parts of the code that we don't get it from the monkeypatch
-    if _type == MODULE:
+    summary_info = {
+        'variables': {},  # Stores mapping of variables and its description, types
```
tbpg, line 386: Missing something after "types" in the comment?

docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
...
+            }
+
+            # Trim the top summary but maintain its formatting.
+            indexes = []
+            for types in var_types:
+                index = summary.find(types)
+                if index > -1:
+                    indexes.append(index)
+
+            # If we found types needing further process, locate its index,
```
tbpg, line 417:
```suggestion
            # If we found types needing further processing, locate its index,
```

docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
...
 
-    # Only add summary to parts of the code that we don't get it from the monkeypatch
-    if _type == MODULE:
+    summary_info = {
+        'variables': {},  # Stores mapping of variables and its description, types
+        'returns': [],    # Stores the return info
+        'exceptions': []  # Stores the exception info
+    }
+
+    # Add extracted summary
```
tbpg, line 391: Is there any of this we can pull out into separate functions? The indentation gets pretty deep here, making it more difficult to follow.

docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
...
+            # Store the top summary separately.
+            if index == 0:
+                top_summary = summary
+            else:
+                top_summary = summary[:index]
+                summary = summary[index:]
 
-    if args or sig:
+                # Clean the string by cleaning newlines and backlashes, then split by white space.
+                parsed_text = " ".join(filter(None, re.split('\n|  |\\\\',summary))).split(" ")
```
tbpg, line 429: Should we handle all forms of whitespace (`\s`)?

tbpg, line 429: Use `r'...'` to avoid escaping issues?

docfx_yaml/extension.py
```diff
@@ -365,19 +382,133 @@ def _update_friendly_package_name(path):
...
+            
+                cur_type = ''
+                words = []
+                arg_name = ''
+                index = 0
+
+                # Using counter iteration to easily extract names rather than
+                # coming up with more complicated stopping logic for each tags.
+                while index <= len(parsed_text):
```
tbpg, line 438: Overall comment: I'm surprised Sphinx doesn't handle this parsing under the hood. If it doesn't and we have to do this, that's fine. But, just commenting to confirm there isn't some sort of already-parsed representation of the comment?

docfx_yaml/extension.py
```diff
@@ -70,7 +70,9 @@ class Bcolors:
 REFMETHOD = 'meth'
 REFFUNCTION = 'func'
 INITPY = '__init__.py'
-REF_PATTERN = ':(py:)?(func|class|meth|mod|ref):`~?[a-zA-Z_\.<> ]*?`'
+REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.<> ]*?`'
```
tbpg, line 73: Overall comment (not just this line): Any tests for this file/these changes?

tbpg (changes requested):

busunkim96 (commented):
It's odd that this needs to be handled separately. The docstrings in spanner are considered "default" by sphinx. The docstrings you see in the generated libraries and most newer libraries are "google style" and are processed through the Napoleon package. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html

dandhlee (commented):
Thanks for the feedback! @busunkim96 I was able to find the napoleon module however since Sphinx stops after having `reStructuredText` I'd still have to take that and convert them to YAML friendly structure, needing this parser. I'll clean the code up in a bit.

dandhlee:
Updated to address the following:
- refactored parser into `_extract_docstring_info` function
- makes use of `GoogleDocstring` function to handle both `reStructuredText` content coming in, as well as Google style docstrings coming in.
  - keeping the sanitizer with `re.split` for further sanitizing the string for parsing purpose
- ignores `cls` entries on parameters which are `self` arguments passed in
  - added coverage for `*kwarg` and `**kwargs` entries
- catch specific types and not generic Exception types
- added test cases to cover for updated regex and one of spanner's functions with various types to process

docfx_yaml/extension.py
tbpg: Are there tests that verify the output?

tbpg (commented):
Overall, looks good. Still more manual parsing than I'd like, but it sounds like it's necessary.

docfx_yaml/extension.py
dandhlee: I've realized the test simply generates content but it doesn't seem to verify the output. Filed #44 to track this, after offline chat moving forward at the moment with generation part and adding back the verification after.

tbpg (approved):

PR #42, opened 2021-06-16
fix: remove function and method name override

dandhlee:

We opted to shorten the names displayed on the functions so that we don't show duplicated name and syntax to the users. The syntax is already captured inside `syntax['content']` field, so no need to duplicate it back into the name field.

- [x] Tests pass


dandhlee:
#22 made this fix for References and other entries where applicable conditions apply (whenever `_create_reference` function is called), while this PR updates for Function and Method types.

tbpg (approved):

Issue #44, opened 2021-06-17
Improve testing infrastructure to verify the output


dandhlee:
Googlers see b/206698957.

 **Is your feature request related to a problem? Please describe.**
Currently the testing infrastructure using `tox -e docs` does not verify the output produced.

 **Describe the solution you'd like**
Whether using nox, tox or other forms of testing tools, have the test verify the output as well so we're not making changes that would be a breaking change.

 **Describe alternatives you've considered**
For now, we're okay to move on as-is to ensure that the generation runs. We should revisit this ASAP to verify the content it outputs.

 **Additional context**
N/A

PR #45, opened 2021-06-18
fix: complete toc disambiguation

dandhlee:

In #16 , it looked for cases which has `google.cloud.package.module` to shorten the entries, and if there are multiple `modules` then we relied on using different package names. However this did not work if there were more parts involved, and if the unique disambiguation we were looking for fell in the middle.

To fix this, for all entries at the same level I've created a map of all words separated by `.` and determine whether there are unique entries of the word or not. It won't matter if there are only single entries of each entries for the left nav. If there are multiple entries, we look for unique entry + name to use in the left nav. Since there is only one unique identifier for each entry, that should be used along with the name to completely disambiguate given a list.

This should now fix all issues with duplicated entries in the left nav!

- [x] Tests pass


dandhlee:
:O neat, the Kokoro job gets linked here if I run based on this PR's git hash :D (it failed because I deleted a blob mid-run for a job, ran another job and it should pass.)

docfx_yaml/extension.py
```diff
@@ -720,23 +720,42 @@ def sanitize_uidname_field(toc_yaml):
                 sanitize_uidname_field(module['items'])
             module.pop('uidname')
 
-    # Parses the package name and returns package name and module name.
-    def find_package_name(package_name):
+    # Parses the package name and returns unique identifer and name.
+    def find_unique_name(package_name, entries):
```
tbpg, line 724: I hate to say, this is begging for a unit test. It appears to be self contained and we know exactly what the output should be based on the given input?

docfx_yaml/extension.py
```diff
@@ -720,23 +720,42 @@ def sanitize_uidname_field(toc_yaml):
...
         for module in toc_yaml:
-            names[module['name']] = 1 if module['name'] not in names else 2
+            module_name = module['name']
+            if module_name not in names:
+                names[module_name] = {}
+
+            # Split the name and mark all duplicates.
+            # There will be at least 1 unique identifer for each name.
+            for part in module['uidname'].split("."):
+                names[module_name][part] = 1 if part not in names[module_name] else 2
```
tbpg, line 745: Should this increment instead of just using 2? I was confused at first.

tbpg (changes requested):

docfx_yaml/extension.py
dandhlee: Could this wait until #44 ? I'm planning on addressing all of this with that.

docfx_yaml/extension.py
dandhlee: Updated. Since the count just has to be greater than 1 and that specific `x = val_1 if condition else val_2` syntax doesn't work with incrementing counter I've just wanted to keep it at 2, but it would be better to increase readability over few lines of code.

docfx_yaml/extension.py
tbpg: I don't think so. This wouldn't be a test for verifying output, it would be a test of this function.

Right now, reviewing this, it seems OK. But, a test would make me feel much better about it.

docfx_yaml/extension.py
```diff
@@ -720,23 +720,45 @@ def sanitize_uidname_field(toc_yaml):
...
+            # For example, if given
+            #   "google.cloud.spanner.v1.params_v1.types"
+            #   "google.cloud.spanner.v1.instance_v1.types"
+            # it will return "instace_v1" or "params_v1" and "types".
+            if name != "google" and name != "cloud" and entries[name] == 1:
                 return [name, package_name[-1]]
 
     # Used to disambiguate names that have same entries.
     def disambiguate_toc_name(toc_yaml):
         names = {}
```
tbpg, line 736: Consider a different name from `names`, as this holds the number of times each name occurs, not just the names. `name_counts`?

tbpg (changes requested):

docfx_yaml/extension.py
dandhlee: Sounds good.

docfx_yaml/extension.py
dandhlee: changed it to `name_entries` since we call its list `entries` above in the `find_package_name` function!

dandhlee:
* Updated to move the functions out to easily have them tested with unittest
* Added `test_unit.py` file for unit tests, as well as some yaml test files
* Added `tox -e unittest` for the GitHub Action to use and test (already reflected in the checks)

Ready for review! Please take a look again :)

tests/test_unit.py
```diff
@@ -0,0 +1,48 @@
...
+class TestGenerate(unittest.TestCase):
+    def test_find_unique_name(self):
+
+        entries = {}
+        
+        # Disambiguate with unique entries.
+        entry1_1 = "google.cloud.aiplatform.v1.schema.predict.instance_v1.types"
+        entry1_2 = "google.cloud.aiplatform.v1beta2.schema.predict.instance_v1.types"
+        want1_1 = "v1.types"
+        want1_2 = "v1beta2.types"
```
tbpg, line 17: Why the `1`? (e.g. entry**1**_1).

tests/test_unit.py
```diff
@@ -0,0 +1,48 @@
...
+        got1_1 = find_unique_name(entry1_1.split("."), entries)
+        got1_2 = find_unique_name(entry1_2.split("."), entries)
+
+        self.assertEqual(want1_1, ".".join(got1_1))
+        self.assertEqual(want1_2, ".".join(got1_2))
+
+
+    def test_disambiguate_toc_name(self):
+
+        want_file = open('tests/yaml_post.yaml', 'r')
```
tbpg, line 35: Can you shorten these test files to be 10s of lines that are easy to understand, rather than ~1100 that are difficult to compare?

tbpg (changes requested):
Thank you!

tests/test_unit.py
dandhlee: I had few more sets of tests, then realized one will be good enough to test but forgot to remove the grouping prefix. Removed it.

tests/test_unit.py
dandhlee: done!

tbpg (approved):
Thank you! And thank you for adding tests. :smile:

PR #55, opened 2021-06-23
fix: update parser to correctly parse desired tokens

### LINKED ISSUE #52: plugin parses docstring differently than expected
With the given docstring:
```python
        """Union[int, None]: Expiration time in milliseconds for a partition.

        If :attr:`partition_expiration` is set and :attr:`type_` is
        not set, :attr:`type_` will default to
        :attr:`~google.cloud.bigquery.table.TimePartitioningType.DAY`.
        """
```
It fails to parse properly and treat the whole syntax as a top line summary when it encounters <xref:type_> since we look for `:type` in the code. This should be prevented by only looking for `:type` and nothing before/after.

dandhlee:

Updated the parser in `_extract_docstring_info` to correctly parse for tokens by specifically looking for strict match like `:type` and not fail on input like `<xref:type_>`. 

As well, updated the extractor to handle different ordering of docstring tokens. While `GoogleDocstring` only returns in specific order, for example `:param` comes before `:type` and `:returns:` comes before `:rtype:` but handwritten libraries sometimes flip these, and I don't see in Google Docstrings page that it should always come in specific order. Returns type is the only set that the extractor trips on different ordering, updated this bit.

Adding unit tests for the above bits (and the function in general) as well.

Fixes #52 

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -270,6 +270,117 @@ def _extract_signature(obj_sig):
     return signature, parameters
 
 
+# Given documentation docstring, parse them into summary_info.
+def _extract_docstring_info(summary_info, summary, name):
+    top_summary = ""
+
+    # Initialize known types needing further processing.
+    var_types = {
+        ':rtype:': 'returns',
+        ':returns:': 'returns',
+        ':type': 'variables',
+        ':param': 'variables',
+        ':raises': 'exceptions',
+        ':raises:': 'exceptions'
+    }
+        
+    # Clean the string by cleaning newlines and backlashes, then split by white space.
+    config = Config(napoleon_use_param=True, napoleon_use_rtype=True)
+    # Convert Google style to reStructuredText
+    parsed_text = str(GoogleDocstring(summary, config))
+
+    # Trim the top summary but maintain its formatting.
+    indexes = []
+    for types in var_types:
+        # Ensure that we look for exactly the string we want.
+        #  Adding the extra space for non-colon ending types
+        #  helps determine if we simply ran into desired occurrence
+        #  or if we ran into a similar looking syntax but shouldn't
+        #  parse upon it.
+        types += ' ' if types[-1] != ':' else ''
+        if types in parsed_text:
+            index = parsed_text.find(types)
+            if index > -1:
+                # For now, skip on parsing custom fields like attribute
+                if types == ':type ' and 'attribute::' in parsed_text:
+                    continue
+                indexes.append(index)
+
+    # If we found types needing further processing, locate its index,
+    # if we found empty array for indexes, stop processing further.
+    index = min(indexes) if indexes else 0
+
+    # Store the top summary separately.
+    if index == 0:
+        top_summary = summary
+    else:
+        top_summary = parsed_text[:index]
+        parsed_text = parsed_text[index:]
+
+        # Clean up whitespace and other characters
+        parsed_text = " ".join(filter(None, re.split(r'\n|  |\|\s', parsed_text))).split(" ")
+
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -270,6 +270,124 @@ def _extract_signature(obj_sig):
...
+    parsed_text = str(GoogleDocstring(summary, config))
+
+    # Trim the top summary but maintain its formatting.
+    indexes = []
+    for types in var_types:
+        # Ensure that we look for exactly the string we want.
+        #  Adding the extra space for non-colon ending types
+        #  helps determine if we simply ran into desired occurrence
+        #  or if we ran into a similar looking syntax but shouldn't
+        #  parse upon it.
```
tbpg, line 299: Minor: extra space at the beginning of these comments?

docfx_yaml/extension.py
```diff
@@ -270,6 +270,124 @@ def _extract_signature(obj_sig):
...
+                    continue
+                indexes.append(index)
+
+    # If we found types needing further processing, locate its index,
+    # if we found empty array for indexes, stop processing further.
+    index = min(indexes) if indexes else 0
+
+    # Store the top summary separately.
+    if index == 0:
+        top_summary = summary
```
tbpg, line 315: Can we return at this point and avoid needing the `else` (and the indentation that comes with it)?

docfx_yaml/extension.py
```diff
@@ -270,6 +270,124 @@ def _extract_signature(obj_sig):
...
+    else:
+        top_summary = parsed_text[:index]
+        parsed_text = parsed_text[index:]
+
+        # Clean up whitespace and other characters
+        parsed_text = " ".join(filter(None, re.split(r'\n|  |\|\s', parsed_text))).split(" ")
```
tbpg, line 321: Missed this before -- why do we need `\n` and ` ` in addition to `\s`?

tbpg: I'm not sure I understand `" ".join(stuff).split(" ")`. Isn't that the same as `stuff`?

tests/test_unit.py
```diff
@@ -44,5 +45,128 @@ def test_disambiguate_toc_name(self):
...
+    AttributeError: if `condition x`.
+"""
+
+        top_summary1_got = _extract_docstring_info(summary_info1_got, summary1, "")
+
+        self.assertEqual(top_summary1_got, top_summary1_want)
+        self.assertEqual(summary_info1_got, summary_info1_want)
+
+
+        ## Test for input coming in mixed format.
```
tbpg, line 104: Consider creating a separate test case for each `summary`. It's a bit hard to follow all of these as-is.

tbpg (changes requested):

docfx_yaml/extension.py
plamut: If `stuff` items contain spaces, the resulting `stuff` is not the same as the original one.

```py
>>> stuff = ["one two", "three four"]
>>> " ".join(stuff).split(" ")
['one', 'two', 'three', 'four']
```

(no idea what the line does, though, not familiar with the extension :smile: )

tests/test_unit.py
dandhlee: done.

dandhlee (commented):
Thanks for the review! Please take a look again :)

docfx_yaml/extension.py
tbpg: Just `return summary` directly?

tbpg (approved):

dandhlee:
done! Updated to return summary directly.

PR #56, opened 2021-06-23
fix: properly handle Raises section for GoogleDocstring

### LINKED ISSUE #53: GoogleDocstring does not parse `Raises` section properly
First encountered in https://github.com/googleapis/google-resumable-media-python/pull/230

##Expected behavior

The docstrings are handled properly by `GoogleDocstring` method

##Actual behavior

When the given type of the exception under `Raises` section is not a simple type but an xref or non-parsed xref given, `GoogleDocstring` throws an `AttributeError` exception. While I cannot find a proper standard regarding this under Google Docstring types, we should handle this properly to ensure that the plugin doesn't run into any issues.

dandhlee:

When there's special characters in 
```
Raises:
    type: description
```
Raises section for type, `GoogleDocstring` does not handle it well and crashes. To avoid this, and to also have it properly be handled by `GoogleDocstring`, I've decided to strip all occurrences of `~type` and `<xref:type>` entries from special characters before running `GoogleDocstring`.

Added a unit test to verify that the strings get parsed properly.

Since this touches on the same function `_extract_docstring_info` as #55, I'll be adding unittest for the function's portion after it's merged and I've resolved git merge conflicts.

Fixes #53 

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -73,6 +73,7 @@ REFMETHOD = 'meth'
 REFFUNCTION = 'func'
 INITPY = '__init__.py'
 REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.<> ]*?`'
+REF_PATTERN_LAST = '~(([a-zA-Z0-9_<>]*\.)*[a-zA-Z0-9_<>]*)'
 
 PROPERTY = 'property'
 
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -498,7 +555,8 @@ def _extract_docstring_info(summary_info, summary):
 
     # Add extracted summary
     if lines != []:
-        lines = _resolve_reference_in_module_summary(lines)
+        for PATTERN in [REF_PATTERN, REF_PATTERN_LAST]:
+            lines = _resolve_reference_in_module_summary(PATTERN, lines)
```
tbpg, line 559: Drop the for loop and call the function directly twice? I was confused at first with what was happening here.

tests/test_unit.py
```diff
@@ -44,5 +47,67 @@ def test_disambiguate_toc_name(self):
...
+Raises:
+    ~google.resumable_media.common.DataCorruption: If the download's
+        checksum doesn't agree with server-computed checksum.
+    ValueError: If the current :class:`Download` has already
+        finished.
+"""
+        lines_got = lines_got.split("\n")
+
+        PATTERNS = [REF_PATTERN, REF_PATTERN_LAST]
+        for PATTERN in PATTERNS:
```
tbpg, line 80: Is the Python style to use all caps for references to constants? Would expect this to be `pattern`.

docfx_yaml/extension.py
```diff
@@ -73,6 +73,7 @@ class Bcolors:
 REFFUNCTION = 'func'
 INITPY = '__init__.py'
 REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.<> ]*?`'
+REF_PATTERN_LAST = '~(([a-zA-Z0-9_<>]*\.)*[a-zA-Z0-9_<>]*)'
```
tbpg, line 76: Add a comment saying what this is for?

tbpg (commented):
A few minor comments.

docfx_yaml/extension.py
dandhlee: done. I'll add comments as well.

tests/test_unit.py
dandhlee: references shouldn't be. Done.

docfx_yaml/extension.py
dandhlee: done.

docfx_yaml/extension.py
```diff
@@ -184,23 +187,31 @@ def _refact_example_in_module_summary(lines):
     return new_lines
 
 
-def _resolve_reference_in_module_summary(lines):
+def _resolve_reference_in_module_summary(PATTERN, lines):
```
tbpg, line 190: Same `PATTERN` vs `pattern` comment from before.

tbpg (approved):

docfx_yaml/extension.py
dandhlee: done.

PR #61, opened 2021-06-25
fix: deduplicate entries in children and references

dandhlee:

There was an issue with the way Sphinx imports docstrings, and processes certain items twice in `process_docstring` function. Instead of checking for duplicate entries in the children/reference, we'll check to see if we've already processed a docstring or not and solve it from the root cause.

Fixes internally filed issue

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -630,15 +630,13 @@ def process_docstring(app, _type, name, obj, options, lines):
     """
 
     # Check if we already processed this docstring.
-    if name not in app.env.docfx_uid_names:
-        # Checking in dictionary("{}") takes O(1) average time for a bit more space,
-        # whereas checking in lists("[]") takes O(n) average time.
-        app.env.docfx_uid_names[name] = ''
-    else:
+    if name in app.env.docfx_uid_names:
         return
 
-    # Use exception as class
+    # Register current docstring to a set.
+    app.env.docfx_uid_names[name] = ''
 
+    # Use exception as class
     if _type == EXCEPTION:
         _type = CLASS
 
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -626,6 +628,15 @@ def process_docstring(app, _type, name, obj, options, lines):
     """
     This function takes the docstring and indexes it into memory.
     """
+
+    # Check if we already processed this docstring.
+    if name not in app.env.docfx_uid_names:
+        # Checking in dictionary("{}") takes O(1) average time for a bit more space,
+        # whereas checking in lists("[]") takes O(n) average time.
```
tbpg, line 635: Not sure we need this, but it doesn't hurt. Using a dict as a set is standard.

docfx_yaml/extension.py
```diff
@@ -626,6 +628,15 @@ def process_docstring(app, _type, name, obj, options, lines):
     """
     This function takes the docstring and indexes it into memory.
     """
+
+    # Check if we already processed this docstring.
+    if name not in app.env.docfx_uid_names:
```
tbpg, line 633: I would flip this, returning early if it's already processed. Then, you don't need the `else`.

docfx_yaml/extension.py
dandhlee: done.

docfx_yaml/extension.py
dandhlee: done!

dandhlee:
Thank you! Ready for review :)

tbpg (approved):

PR #69, opened 2021-07-13
chore: allow testing plugin without running nox

### LINKED ISSUE #68: Allow testing the plugin through the Kokoro job without running through Nox sessions
The current Kokoro job that runs `sphinx-build` for all repositories relies on using Nox, as we also use this job to refresh the docs. We should be able to test the plugin (and perhaps Sphinx as well) with newer versions without much needed manual work or a release having to be done.

We could also do [pre-releases](https://www.python.org/dev/peps/pep-0440/#pre-releases) as Bu Sun helped point when needed.

dandhlee:

Minor update to the testing script for the plugin for running the plugin across all Python client libraries. If `TEST_PLUGIN=true` is passed on, it will simply run `sphinx-build` with the default sphinx installed based on the version of the hash passed in, allowing us to test the newer versions of the plugin more effectively. We could also make this as part of a presubmit test when needed (expected to take ~30 minutes to run).

When the environmental variable above is passed, it will not update any docs.

Fixes #68 

- [x] Tests pass

### DIFF
--- a/.kokoro/generate-docs.sh
+++ b/.kokoro/generate-docs.sh
@@ -30,6 +30,8 @@ gcloud auth activate-service-account --key-file ${GOOGLE_APPLICATION_CREDENTIALS
 # Install dependencies.
 python3 -m pip install --user --quiet --upgrade nox
 python3 -m pip install --user gcp-docuploader
+python3 -m pip install -e .
+python3 -m pip install recommonmark
 
 # Retrieve unique repositories to regenerate the YAML with.
 for bucket_item in $(gsutil ls 'gs://docs-staging-v2/docfx-python*' | sort -u -t- -k5,5); do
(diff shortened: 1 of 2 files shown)


.kokoro/generate-docs.sh
```diff
@@ -93,6 +95,12 @@ for bucket_item in $(gsutil ls 'gs://docs-staging-v2/docfx-python*' | sort -u -t
     ## upload docs
     # python3 -m docuploader upload docs/_build/html --metadata-file docs.metadata --staging-bucket "${STAGING_BUCKET}"
 
+    # Test running with the plugin version locally.
+    if [[ ${TEST_PLUGIN} == "true" ]]; then
```
tbpg, line 99: I think you need quotes around this in case it's not set?
```suggestion
    if [[ "${TEST_PLUGIN}" == "true" ]]; then
```

dandhlee:
Updated, please take a look again!

tbpg (approved):

PR #72, opened 2021-07-20
fix: disambiguate all entry names to clarify duplicate names

dandhlee:

Using the de-duplicated names from the `disambiguate_toc_name` function, updating entries of all disambiguated names other than just the TOC file. This includes entries on Overview pages as well as for references.

Instead of outputting the file content each time `yaml_data` is pulled, wait until names are disambiguated. The function is updated to keep track of all names that are disambiguated. 

For each of the yaml data entry and reference, cross-check with the names that were disambiguated and make necessary updates. Then, with the updated names, output the file content at the end.

There is a minor side effect of longer entry names, for example from `client` to `bigquery_storage_v1.client` and possibly longer if necessary; but I think we can live with slightly longer name for clarity.

Updates internally filed issue

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -809,8 +809,12 @@ def find_unique_name(package_name, entries):
     return [package_name[-1]]
 
 # Used to disambiguate names that have same entries.
+# Returns a dictionary of names that are disambiguated in the form of:
+# {uidname: disambiguated_name}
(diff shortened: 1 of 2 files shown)


dandhlee:
Confirming the output looks good using `bigquery-storage` and `aiplatform` first.

dandhlee:
Verified that it looks good!

tbpg (approved):

PR #77, opened 2021-07-24
feat: add subclasses to children and reference

dandhlee:

The function `insert_children_on_class` only cared about underlying class methods, attributes and properties, which left out subclasses of a class. 

To make that happen, search for the parent class by using `datam['module']` and see if there is a parent class that exists, then check that both types are `class` so we don't accidentally insert anything else when we retrieve classes from `app.env.docfx_yaml_classes.get(datam[MODULE])`. 

When actually inserting the children, we separate whether it's class items we're looking for such as attributes, properties and methods, or if it is a subclass of a class with the nested if condition.

Fixes internally filed issue.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -767,7 +767,7 @@ def insert_children_on_module(app, _type, datam):
                         datam['references'].append(_create_reference(obj, parent=module))
                         break
 
- 
+
 def insert_children_on_class(app, _type, datam):
     """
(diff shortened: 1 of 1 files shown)


dandhlee:
Adding this as part of unittest is a bit hard to do because of `app.env.docfx_yaml_classes`, I'll handle the test case as part of testing with goldens in #44.

tbpg (approved):

PR #78, opened 2021-07-25
feat: process xrefs properly

dandhlee:

While resolving the summary, now it fully handles xrefs by shortening references from `google.cloud.long.product.name` to `<xref uid="uid">name</xref>` to allow xrefs to be used in devsite.

After handling that, `_extract_docstring_info` needed to be updated to 
* no longer care about `~xrefs` type, and only `<xref:name>` type
* handle duplicates carefully, and only handle xrefs in non-top_summary part

Then extract xrefs as needed, for now we'll only extract xrefs for Cloud libraries.

It's quite hard to put other product xrefs into `devsite://lang/package` as python package names can heavily differ and is hard to extract from the UID of the xref at this context. For now, I'll be storing them in full UID and handle it in doc-pipeline where I can more easily figure out package based on the UID. I'm tempted to store this in the plugin (`map<uid: package>`), but I'll have to think harder on that. Any suggestions?

Fixes internally filed issue.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -102,6 +102,8 @@ def build_init(app):
     # This stores uidnames of docstrings already parsed
     app.env.docfx_uid_names = {}
 
+    app.env.docfx_xrefs = {}
+
     remote = getoutput('git remote -v')
 
     try:
@@ -190,7 +192,7 @@ def _refact_example_in_module_summary(lines):
 
 
 def _resolve_reference_in_module_summary(pattern, lines):
-    new_lines = []
+    new_lines, xrefs = [], []
     for line in lines:
         matched_objs = list(re.finditer(pattern, line))
         new_line = line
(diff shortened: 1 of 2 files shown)


tbpg (changes requested):
How would the package name extraction work in doc-pipeline?

dandhlee:
@tbpg I was thinking something along the lines of given the uid in xref.yml, we'd look through the `xrefs/python-{package}*.tar.gz.yml` files, but taking a look again seems like there's no good way to process the package without having some sort of an easy access map that's pre-processed.

The only place we can really find the package name is in `docs.metadata`, for example when we see:
```
distribution_name: "google-cloud-bigquery-storage"
name: "bigquerystorage"
```
where `name` corresponds to the package name we want, and `distribution_name` is the closest thing we have to identifying between the uid/product and the package name.

I think we should follow this up offline, I'll come up with a better concrete plan and follow up with it.

dandhlee:
Interestingly, after taking a look at ~56 packages there doesn't seem to be any libraries that cross reference other `google.cloud` libraries. I think it's safe to leave that out for now, and worry about it when we add in non-Cloud libraries.

dandhlee:
Filed #82 for handling cross referencing for other Cloud libraries, which should also allow us to handle non-Cloud libraries as well at that point.

Otherwise, ready for review! :)

tbpg (approved):

PR #85, opened 2021-07-29
fix: handle more xrefs and keep long uid

### LINKED ISSUE #84: Handle more cross references and non-resolvable ones more gracefully
Currently the following xrefs _should_ be resolved but get ignored:

```google.cloud.product_v1.func()```

because of the parenthesis in the end. 

As well, non-resolved cross references get simplified, which sometimes makes it confusing to see what they were. Reinforcing them in code blocks with `` should be good but we shouldn't simplify the original, stripping the useless xref notations should be fine however.

### PR DESCRIPTION
**Pull request by dandhlee**

Editing the regex to look for REF_PATTERN to include parenthesis; however I don't think we should be expecting any cross references that extend beyond the parenthesis. That would mess up trying to find its UID as well, and I don't think that would work naturally either way (How would they show up in its original place?)

When we encounter non-Cloud and third party references, we've been simplifying them which could make them confusing to see. They will still be emphasized by being put in ``, without the additional xref identifiers like `:func:` and other non-alphanumeric characters.

Updated with a new unittest case to reflect upon the changes.

Fixes #84 

- [x] Tests pass

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -136,6 +136,50 @@ Raises:
         self.assertEqual(lines_got, lines_want)
 
 
+    # Test for added xref coverage and third party xrefs staying as-is
+    def test_reference_in_summary_more_xrefs(self):
+        lines_got = """
+If a ~dateutil.time.stream() is attached to this download, then the downloaded
+resource will be written to the stream.
+
+Args:
+    transport (~google.cloud.requests.Session()): A ``requests`` object which can
+        make authenticated requests.
+
+    timeout (Optional[Union[float, Tuple[float, float]]]):
+        The number of seconds to wait for the server response.
+        Depending on the retry strategy, a request may be repeated
+        several times using the same timeout each time.
+
+        Can also be passed as a :func:`~google.cloud.requests.tuple()` (connect_timeout, read_timeout).
+        See :meth:`google.cloud.requests.Session.request()` documentation for details.
+"""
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_unit.py [hunk 1]
```diff
@@ -136,6 +136,50 @@ def test_reference_in_summary(self):
...
+        Depending on the retry strategy, a request may be repeated
+        several times using the same timeout each time.
+
+        Can also be passed as a :func:`~google.cloud.requests.tuple()` (connect_timeout, read_timeout).
+        See :meth:`google.cloud.requests.Session.request()` documentation for details.
+"""
+        lines_got = lines_got.split("\n")
+        # Resolve over different regular expressions for different types of reference patterns.
+        lines_got, xrefs = _resolve_reference_in_module_summary(REF_PATTERN, lines_got)
+        lines_got, xrefs = _resolve_reference_in_module_summary(REF_PATTERN_LAST, lines_got)
```
tbpg, line 160: Anything to check for `xrefs`?

tbpg (approved):
One test comment (thank you for writing one!)

tests/test_unit.py
dandhlee: Yes, we should check them. I'll add a test to check for `google.cloud` uids.

dandhlee:
Updated to check for list of xrefs using `assertCountEqual`, sort of misleading name but compares the two lists we need.

PR #90, opened 2021-07-29
fix: parse xrefs differently with new xref format

### LINKED ISSUE #89: GoogleDocstring does not handle new xref format well
Previously, when xrefs were included for GoogleDocstring it crashed, not knowing how to process it. To work around this, I've took xrefs out of the equation, ran GoogleDocstring, then added them back in. Tedious, but worked. Also included test cases.

However, with the new format of xrefs with `<xref uid="uid">text</xref>`, this was not considered as the previous grep for "xref:" were the only ones included in the unit test, and not with the new format, thus it hasn't worked out well.

Interestingly enough, I'm not sure why it took ~60 repos to find that this would fail, rather than it failing before...

dandhlee:

Updating xref parser to accommodate the new and only xref format: `<xref uid="uid">text</xref>`.

We need to be careful on two parts:
* 1) Carefully extracting only `text` portion to replace so GoogleDocstring doesn't complain on the leftover xref syntax
* 2) Carefully parse through the docstring as we strip whitespace as well, handled by looking at one token ahead when necessary. 

Fixes #89 
Fixes https://github.com/googleapis/python-firestore/issues/408

- [ ] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -460,12 +460,13 @@ def _extract_docstring_info(summary_info, summary, name):
                 elif ':raises' in cur_type and 'xref' in parsed_text[index]:
                     arg_name = f'{parsed_text[index]} {parsed_text[index+1][:-1]}'
                     index += 1
-                # Other forms will not be processed properly, likely a formatting issue.
-                else:
-                    raise ValueError(f"Encountered wrong formatting, please check docstring for {name}")
-                # Initialize empty dictionary if it doesn't exist already
-                if arg_name not in summary_info[var_types[cur_type]] and ':raises' not in cur_type:
-                    summary_info[var_types[cur_type]][arg_name] = {}
+
+                try:
+                    # Initialize empty dictionary if it doesn't exist already
+                    if arg_name not in summary_info[var_types[cur_type]] and ':raises' not in cur_type:
+                        summary_info[var_types[cur_type]][arg_name] = {}
+                except KeyError:
+                    raise KeyError(f"Encountered wrong formatting, please check docstring for {name}")
 
             # Empty target string
             words = []
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -335,16 +335,18 @@ def _extract_docstring_info(summary_info, summary, name):
 
         summary_part = parsed_text[initial_index:]
        
-        # Remove all occurrences of "<xref:type>"
+        # Remove all occurrences of "<xref uid="uid">text</xref>"
         while "<xref" in summary_part:
 
-            # Expecting format of "<xref:type>:"
+            # Expecting format of "<xref uid="uid">text</xref>"
             if "<xref" in summary_part:
                 initial_index += summary_part.find("<xref")
                 original_type = parsed_text[initial_index:initial_index+(parsed_text[initial_index:].find('/xref>'))+6]
                 initial_index += len(original_type)
                 original_type = " ".join(filter(None, re.split(r'\n|  |\|\s|\t', original_type)))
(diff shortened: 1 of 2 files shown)


dandhlee:
Testing on Kokoro with `TEST_PLUGIN=true` option first

docfx_yaml/extension.py
```diff
@@ -329,22 +329,24 @@ def _extract_docstring_info(summary_info, summary, name):
...
+        # Remove all occurrences of "<xref uid="uid">text</xref>"
+        while "<xref" in summary_part:
 
-            # Expecting format of "<xref:type>:"
-            if "<xref:" in summary_part:
+            # Expecting format of "<xref uid="uid">text</xref>"
+            if "<xref" in summary_part:
                 initial_index += summary_part.find("<xref")
-                original_type = parsed_text[initial_index:initial_index+(parsed_text[initial_index:].find('>'))+1]
+                original_type = parsed_text[initial_index:initial_index+(parsed_text[initial_index:].find('/xref>'))+6]
```
tbpg, line 344: What is 6? Can we make that a constant or something?

docfx_yaml/extension.py
```diff
@@ -329,22 +329,24 @@ def _extract_docstring_info(summary_info, summary, name):
...
                 initial_index += len(original_type)
                 original_type = " ".join(filter(None, re.split(r'\n|  |\|\s|\t', original_type)))
-                safe_type = 'xref_' + original_type[6:-1]
+                # Extract text from "<xref uid="uid">text</xref>"
```
tbpg, line 347: It's a bit late now, but we could also add this UID to the `references` section with however we want it to show up. Then, we wouldn't need to do any HTML parsing.

docfx_yaml/extension.py
```diff
@@ -451,10 +453,20 @@ def _extract_docstring_info(summary_info, summary, name):
...
+                elif ':raises' in cur_type and 'xref' in parsed_text[index]:
+                    arg_name = f'{parsed_text[index]} {parsed_text[index+1][:-1]}'
+                    index += 1
+
+                try:
+                    # Initialize empty dictionary if it doesn't exist already
+                    if arg_name not in summary_info[var_types[cur_type]] and ':raises' not in cur_type:
+                        summary_info[var_types[cur_type]][arg_name] = {}
+                except KeyError:
+                    raise KeyError(f"Encountered wrong formatting, please check docstring for {name}")
```
tbpg, line 469: I'm not sure this will work. What if there is a release with bad docs? We still need to generate docs for it. So, we should make our best guess and move on.

tbpg (changes requested):

docfx_yaml/extension.py

docfx_yaml/extension.py
dandhlee: I'll keep that in mind perhaps for non-Cloud or third party xrefs.

docfx_yaml/extension.py
dandhlee: Repos should have a docs-presubmit which runs the docs job to see whether it successfully runs this plugin on the repo for `sphinx-build` or not, for example see https://github.com/googleapis/python-firestore/pull/407

While it's usually not set to be a required check, this lets contributors know that docs job would fail or pass before triggering a release.

tbpg: Sounds good. Might need to update this depending on how back filling old libraries goes.

tbpg (approved):

docfx_yaml/extension.py
dandhlee: I do have the Kokoro job set up that could try and run it for all versions, I'll try testing on that for older versions.

PR #93, opened 2021-08-12
fix: recover lost function arguments and argument types

dandhlee:

When there is a `*` argument present in the function signature, they are not included in `args` from `inspect.getfullargspec` function. Iterating through the entire list to make sure everything has been counted for will ensure all variables are accounted for.

As well, when there isn't a docstring entry for functions they only had their names (`id`) included in the table of parameters. Recovering their types if specified in the function signature, otherwise they'll be removed to have the tables not look so empty.

Fixes internally filed issue.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -553,12 +553,15 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
                 # Attempt to add default values to arguments.
                 try:
                     for count, default in enumerate(argspec.defaults):
-                        cut_count = len(argspec.defaults)
+                        # Find the first index which default arguments start at.
+                        # Every argument after this offset_count all have default values.
+                        offset_count = len(argspec.defaults)
                         # Only add defaultValue when str(default) doesn't contain object address string(object at 0x)
                         # inspect.getargspec method will return wrong defaults which contain object address for some default values, like sys.stdout
-                        # Match the defaults with the count
+                        # Find the index of the current default value argument
+                        index = len(args) + count - offset_count
                         if 'object at 0x' not in str(default):
-                            args[len(args) - cut_count + count]['defaultValue'] = str(default)
+                            args[index]['defaultValue'] = str(default)
                 # If we cannot find the argument, it is missing a type and was taken out intentionally.
                 except IndexError:
                     pass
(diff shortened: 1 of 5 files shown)


dandhlee:
Ran a test Kokoro job, seems to work fine but it failed on `python-bigquery-sqlalchemy` which is expected - it currently produces no documentation for DocFX YAML.

docfx_yaml/extension.py
```diff
@@ -517,22 +517,51 @@ def _update_friendly_package_name(path):
...
-                        args[len(args) - cut_count + count]['defaultValue'] = str(default)
+                # Attempt to add default values to arguments.
+                try:
+                    for count, default in enumerate(argspec.defaults):
+                        cut_count = len(argspec.defaults)
+                        # Only add defaultValue when str(default) doesn't contain object address string(object at 0x)
+                        # inspect.getargspec method will return wrong defaults which contain object address for some default values, like sys.stdout
+                        # Match the defaults with the count
+                        if 'object at 0x' not in str(default):
+                            args[len(args) - cut_count + count]['defaultValue'] = str(default)
```
tbpg, line 561: I don't understand `len(args) - cut_count + count`. Can you clarify in a comment, different variable names, or something else?

docfx_yaml/extension.py
```diff
@@ -643,12 +672,26 @@ def _update_friendly_package_name(path):
...
         for arg in args:
+            arg_id.append(arg['id'])
+
             if arg['id'] in variables:
                 # Retrieve argument info from extracted map of variable info
                 arg_var = variables[arg['id']]
                 arg['var_type'] = arg_var.get('var_type') if arg_var.get('var_type') else ''
                 arg['description'] = arg_var.get('description') if arg_var.get('description') else ''
+
+        # Add any variables we might have missed from extraction.
```
tbpg, line 685: Do we need to worry about order of arguments?

docfx_yaml/extension.py
```diff
@@ -643,12 +672,26 @@ def _update_friendly_package_name(path):
...
+        for variable in variables:
```
tbpg, line 686: Could we have a single loop over `variables` and not `args`?

tbpg (changes requested):
Any tests? :smile:

dandhlee (commented):
Hard to make a specific unit test for the added bits as they need to be part of Sphinx with `app`, it would be easier to have this with the goldens, planning to fix it in #44.

tbpg (approved):

PR #100, opened 2021-08-18
fix: retrieve file name as much as possible 

### LINKED ISSUE #99: Some object types do not have filepath associated to it in datam
`inspect.getsourcefile(obj)` or `inspect.getfile(obj)` does not work for built-in module, class, or function. Some types get treated as a built-in type during Sphinx, which now fails when we try to grab TOC entry names based on its file path from #97. There's no easy way to retrieve file path names for these at this point in the plugin, unless we somehow change the way `inspect` library works.

dandhlee:
Some entries are missing `path: .*\.py` entries in their metadata, which makes the change from #97 crash on those entries. We will try to see if the class associated to the missing item exists in `app.env.docfx_class_filepaths` and parse using that filepath and disambiguate if needed. If the filepath cannot be retrieved, fall back on using `node_name`.

Updating the disambiguate logic as #97 had a side effect of making the names ambiguous if there are multiple entries from the same file, which now needs disambiguation and returns `disambiguated_name.disambiguated_name` which is redundant. 

Fixes #99 

- [x] Tests pass, Kokoro failure is unrelated.

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -101,6 +101,8 @@ def build_init(app):
     app.env.docfx_info_uid_types = {}
     # This stores uidnames of docstrings already parsed
     app.env.docfx_uid_names = {}
+    # This stores file path for class when inspect cannot retrieve file path
+    app.env.docfx_class_paths = {}
 
     app.env.docfx_xrefs = {}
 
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -916,7 +920,8 @@ def find_unique_name(package_name, entries):
         #   "google.cloud.spanner.v1.instance_v1.types"
         # it will return "instace_v1" or "params_v1" and "types".
         if name != "google" and name != "cloud" and entries[name] == 1:
-            return [name, package_name[-1]]
+            # Return only what's necessary, do not return duplicate parts twice.
+            return [name, package_name[-1]] if name != package_name[-1] else [name]
```
tbpg, line 924: Perhaps check this condition in the `if` a line above then let the default return handle it a couple lines down?

The comment is nice about why you have that particular check.

tbpg (approved):
One optional comment, otherwise LGTM.

docfx_yaml/extension.py
dandhlee: Done.

PR #101, opened 2021-08-19
chore: add yoshi-python as addititional owner

dandhlee:
Onboarded Yoshi team with the plugin!


tbpg (approved):
Please also give the team write access so their approvals "count" on PRs.

PR #104, opened 2021-08-21
fix: use the uid for toc entries

### LINKED ISSUE #103: Some entries in the TOC use wrong filename
After #97 and #100, there are still more side effects of using the filename where the file path exists but uses the wrong desired name. For example, given 
```
google/cloud/bigquery/dbapi/connection.py
```
if `connection.py` entry is encountered first, the group name becomes `connection` despite its parent entry `dbapi` existing. It should correctly use `dbapi` in this case, instead of `connection` as its name in the entry.

dandhlee:
#100 was submitted where we failed to retrieve the file name. However, if the file name exists and we have entries entered into the TOC before its proper parent entry as shown in the example in #103, then the entry names get mixed up in the TOC.

I've retrieved the UID for the parent entries, so we can easily check against the existing UID. This should now cover almost every case unless there are very unique circumstances that I'm not aware of. With the given existing UID, we should be able to safely retrieve the names without doing much more work other than disambiguating if needed.

Fixes #103.

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -983,7 +983,7 @@ def group_by_package(toc_yaml):
     for package_group in package_groups:
         new_toc_yaml.append(package_groups[package_group])
 
-    return  new_toc_yaml
+    return new_toc_yaml
 
 
 # Given the full uid, return the package group including its prefix.
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -983,23 +983,7 @@ def group_by_package(toc_yaml):
     for package_group in package_groups:
         new_toc_yaml.append(package_groups[package_group])
 
-    return new_toc_yaml
-
-
-# Checks names against their uid, if they're different update it.
-def check_name_with_uid(toc_yaml):
-    for entry in toc_yaml:
-        if entry.get("items"):
-            check_name_with_uid(entry['items'])
-
-        # Don't update the name for overview pages or entries without uid
-        if entry['name'] == "Overview" or not entry.get("uid"):
-            continue
-
-        # If there is a discrepancy in the name, use one from uid.
-        name_from_uid = entry['uid'].split(".")[-1]
-        if entry['name'] != name_from_uid:
-            entry['name'] = name_from_uid
+    return  new_toc_yaml
 
 
 # Given the full uid, return the package group including its prefix.
@@ -1202,13 +1186,13 @@ def build_finished(app, exception):
                         del(obj['source'])
 
             yaml_map[uid] = [yaml_data, references]
-           
+
             # Parse the name of the object.
             # Some types will need additional parsing to de-duplicate their names and contain
-            # a portion of their parent name for better disambiguation. This is done in 
+            # a portion of their parent name for better disambiguation. This is done in
             # disambiguate_toc_name
-            
-            node_name = obj.get('class').split(".")[-1] if obj.get('class') else obj['name']
+
+            node_name = uid.split(".")[-1]
 
             # Build nested TOC
             if uid.count('.') >= 1:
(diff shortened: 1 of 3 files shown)


dandhlee:
Testing with Kokoro, it will fail unless #102 is merged in, but seeing if this crashes on any other repo.

dandhlee:
Kokoro was able to run until the expected failure because of #102 not being merged in yet.

docfx_yaml/extension.py
```diff
@@ -986,6 +986,22 @@ def group_by_package(toc_yaml):
...
+            check_name_with_uid(entry['items'])
+
+        # Don't update the name for overview pages or entries without uid
+        if entry['name'] == "Overview" or not entry.get("uid"):
+            continue
+
+        # If there is a discrepancy in the name, use one from uid.
+        name_from_uid = entry['uid'].split(".")[-1]
+        if entry['name'] != name_from_uid:
+            entry['name'] = name_from_uid
```
tbpg, line 1002: Overall question: why can't we generate `entry` with the right `name` in the first place? I would assume we know the `uid` when we generate the `name`?

tests/test_unit.py
```diff
@@ -538,5 +539,82 @@ def test_group_by_package(self):
...
+    def test_check_name_with_uid(self):
+        # Test that none of these entries change.
+        toc_yaml_want = [
+            {
+                "name":"Overview",
+                "uid":"google.cloud.spanner.types"
+            },
+            {
+                "name":"types",
+                "uid":"google.cloud.spanner.types"
```
tbpg, line 552: Two TOC entries with the same `uid`?

docfx_yaml/extension.py
```diff
@@ -1238,6 +1253,9 @@ def convert_module_to_package_if_needed(obj):
     if len(toc_yaml) == 0:
         raise RuntimeError("No documentation for this module.")
 
+    # Ensure names align with the last bits of the uid
+    check_name_with_uid(toc_yaml)
```
tbpg, line 1257: We're mixing API types. `group_by_package` returns a value and `check_name_with_uid` modifies the value directly. We should pick one.

tbpg (changes requested):
Thank you for including tests!

docfx_yaml/extension.py
dandhlee: Ah, I think that was from the side effect of removing `uid` field on the top level entries. This was the 
```
found_node.pop('uid', 'No uid found')
```

line that I removed to ensure `uid` is now always available. Because it wasn't available we had to use this workaround. Now we should be good to use just the `uid` and not have to do all this extra work.

dandhlee: This should now be avoided that I've restored `uid` fields on parent entries that's missing `uid` from removing the line below on L1203. I'll update the code accordingly. Thanks for catching this!

docfx_yaml/extension.py
dandhlee: Noted. I'll stick with modifying the value directly rather than returning the toc.

tests/test_unit.py
dandhlee: interestingly, they do appear duplicated in the actual toc:
```
    - items:
      - name: Overview
        uid: google.cloud.logging_v2.client
      - name: Client
        uid: google.cloud.logging_v2.client.Client
      name: client
      uid: google.cloud.logging_v2.client
```

I'll update this test case however to avoid confusion.

docfx_yaml/extension.py
dandhlee: Correction, I'll return the value and remove `check_name_with_uid`.

dandhlee:
Please take a look again!

dandhlee:
Looks like OwlBot is having a bad day 😓

parthea (approved):

busunkim96 (approved):

tbpg (approved):

PR #111, opened 2021-08-24
fix: parse markdown header more carefully

### LINKED ISSUE #110: Process markdown headers correctly
The current logic for checking whether a markdown file has a header or not is not sufficient enough. Also, we might have to check to support the other type of headers using `======` blocks.

dandhlee:
There are some files that don't have a header for markdown files and some headers that were wrongly extracted. Also adding support for alternate header type in markdown, see https://www.markdownguide.org/basic-syntax/#headings.

- For the traditional markdown header using `#` for h1 header, check that:
  - [x] There's only one `#` for an h1 header
  - [x] There's exactly one space after one `#`
  - [x] Is not just a short string consisting only of `#`

- For the alternate markdown header using `=` characters below the header, check that:
  - [x] The header above the `=` divider isn't empty
  - [x] The divider line only consists of `=` and whitespace characters if necessary

Otherwise, I've defaulted to returning the file name if no header can be found. I'll leave it to the library to better format the markdown files rather than dealing with individual formatting issues on the plugin.

Updated unit tests. 

Fixes #110.
- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1040,12 +1040,13 @@ def pretty_package_name(package_group):
 
 
 # Check is the current lines conform to markdown header format.
-def is_markdown_header(header_line, prev_line):
-    if "#" in header_line:
+def parse_markdown_header(header_line, prev_line):
+    # Markdown h1 prefix should have only 1 of '#' character followed by exactly one space.
+    h1_header_prefix = "# "
+    if h1_header_prefix in header_line and header_line.count("#") == 1:
         # Check for proper h1 header formatting, ensure there's more than just
-        # the hashtag character.
-        if header_line[header_line.index("#")+1] == " " and \
-            not header_line[header_line.index("#")+2].isspace() and \
+        # the hashtag character, and exactly only one space after the hashtag.
+        if not header_line[header_line.index(h1_header_prefix)+2].isspace() and \
             len(header_line) > 2:
 
             return header_line.strip("#").strip()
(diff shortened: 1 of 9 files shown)


tests/test_unit.py
```diff
@@ -551,28 +627,72 @@ def test_extract_header_from_markdown(self):
...
 
+        # Check that only h1 headers are parsed.
+        header_line_want = "Markdown_example_h2"
+
+        mdfile = open('tests/markdown_example_h2.md', 'r')
+        header_line_got = extract_header_from_markdown(mdfile)
+
+        self.assertEqual(header_line_want, header_line_got)
+
+        mdfile.close()
```
parthea, line 685: This line can be removed if `mdfile = open('tests/markdown_example_h2.md', 'r')` is changed to a with statement. For example, `with open('tests/markdown_example_h2.md', 'r') as mdfile:`

docfx_yaml/extension.py
```diff
@@ -1032,18 +1032,47 @@ def pretty_package_name(package_group):
     return " ".join(capitalized_name)
 
 
+# Check is the current lines conform to markdown header format.
+def is_markdown_header(header_line, prev_line):
+    if "#" in header_line:
+        # Check for proper h1 header formatting, ensure there's more than just
+        # the hashtag character.
+        if header_line[header_line.index("#")+1] == " " and \
```
parthea, line 1040: If we're looking for both header and a space, we could change the check to `if "# " in header_line:` to look for a header and a space

parthea (approved):
Added minor observations, otherwise LGTM.

tests/test_unit.py
dandhlee: Thank you for the tip! Updating the files.

dandhlee:
Updated function name from `is_markdown_header` to `parse_markdown_header` to be more in line with what it returns, which isn't a boolean.

PR #116, opened 2021-08-25
feat: handle additional docstring items

dandhlee:
Adding support for handling docstring items in `top_summary`. Notably, `.. code-block::` and `.. attributes::` will be supported, and any other items as they appear.

Similar to how return/exception/params were extracted, iterating through the summary to look for keywords like `code-block`, `attribute` and more in patterns that show up in `.. keyword::` format. If the docstring is malformed from the source, we do not handle it in the plugin. Hand-written docstrings may not work well against the autogenerated reStructuredText, however we should fix those from the client library side than to make special adjustments in the plugin like the bad formatted keywords.

For each keyword:
- For `code-block`, convert the code snippet in code block using multi-line commenting blocks. 
    - code should be indented, extract lines until there is no indentation/leading whitespace found and/or found other keywords.
    - Adjust spacing and line breaks as necessary to better display them on cloud site.

- For `attributes`, there are 3 main parts: name, description and type.
    - Extract the name first, followed by the description, then type. Once all 3 are extracted, add it to the list of attributes.
    - Attributes will be added to the YAML file directly, to be processed in doc-templates.
    - If attributes are not well formatted, do not add them to the list to be outputted to YAML.

Afterwards, incorporate them back into `top_summary` and to be added into the YAMLs.

Fixes internally filed issue.
- [x] Tests pass


tbpg (changes requested):

dandhlee (commented):
Please take a look again!

tbpg (approved):

PR #123, opened 2021-08-27
fix: make plugin more verbose

### LINKED ISSUE #122: The plugin doesn't show enough docstrings to help debug
When the plugin fails to parse docstring, sometimes it only shows only the relevant part that's broken, and not the bigger parts to help locate where it may be.

dandhlee:
When folks are debugging it's hard to see which part of the docstring there's an error, if they're given only a short snippet. To help track down where the issue originates from, while the original docstring is lost, I've tried to include as much information as possible to help track down where the broken docstring lives in.

Example of additional output with the updated format:
```
ValueError: In google/cloud/bigquery/job/copy_.py
For module google.cloud.bigquery.job, type method:
Failed to parse docstring on google.cloud.bigquery.job.CopyJob.from_api_repr.
```

Fixes #122.

CC @jimfulton

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -836,12 +836,12 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
             try:
                 datam['summary'], datam['attributes'] = _parse_docstring_summary(top_summary)
             except ValueError:
-                debug_line = ""
+                debug_line = []
                 if path:
-                    debug_line += f"In {path}\n"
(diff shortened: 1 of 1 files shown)


parthea (approved):

dandhlee:
string append using `+=` can be quadratic runtime, where as string append using an array and `"".join(string_array)` is linear run time. 😮 

See https://google.github.io/styleguide/pyguide.html#310-strings

Issue #82, opened 2021-07-28
Handle cross references for other libraries and non-Cloud libraries.


dandhlee:
Googlers see b/195685296.

With #78 we will be able to handle cross references that are contained for within a library. However since there are no cross references to another Cloud library, scoping handling of other libraries and other non-Cloud libraries in this issue.

If there are Cloud libraries that cross reference to other Cloud libraries, first phase will be to handle that.

When we add in support for non-Cloud libraries, we will need to handle those as well.

PR #138, opened 2021-09-25
feat: find more items to cross reference

dandhlee:
Given the full UIDs found within syntax, attributes, parameters, returns and exception info, tries to parse through the content and find entries to cross reference to.

After collecting a list of UIDs, we check in specific parts of the YAML that's applicable. I've decided to look for the expanded full UID (i.e. `google.cloud.bigquery_storage_v1.types.SplitReadStreamResponse`) rather than looking for the short version as well (i.e. `SplitReadStreamResponse`) because some of the shorter names shorten to common words like `open` and `type` and there is just not an easy way to determine correct parsing of those. The words don't always just appear as broken by whitespace, but could be contained in non-alphanumeric characters like `[google.cloud.bigquery_storage_v1.types]` and `` `google.cloud.bigquery_storage_v1.types` ``, hence using `keyword in content` approach.

This will fully cover for missing cross references for expanded UIDs and those who can be linked to within the package. For Googlers, see example through `bigquery_storage` on internal site.

Towards #82 for cross reference within the same package.

Fixes b/195685296.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -85,6 +85,7 @@ REF_PATTERN_LAST = '~([a-zA-Z0-9_<>]*\.)*[a-zA-Z0-9_<>]*(\(\))?'
 PROPERTY = 'property'
 CODEBLOCK = "code-block"
 CODE = "code"
+PACKAGE = "package"
 
 
 # Run sphinx-build with Markdown builder in the plugin.
@@ -673,7 +674,7 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
 
     if lines is None:
         lines = []
-    short_name = name.split('.')[-1]
+    short_name = name.split(".")[-1]
     args = []
     # Check how many arguments are present in the function.
     arg_count = 0
@@ -1260,6 +1261,24 @@ def find_markdown_pages(app, outdir):
             })
 
 
+# Looks for cross references to look for in given content, except for
+# current_name.
+def find_cross_references(content, current_name, keyword_map):
+    words = content.split(" ")
+    # Using counter to check if the entry is already a cross reference.
+    for i in range(0, len(words)):
+        for keyword in sorted(keyword_map.keys(), reverse=True):
+            if keyword != current_name and keyword not in current_name and keyword in words[i]:
+                # If it is already processed as cross reference, skip over it.
+                if "<xref" in words[i-1]:
+                    continue
+                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
+                words[i] = words[i].replace(keyword, cross_reference)
+                print(f"Converted {keyword} into cross reference in: \n{content}")
+
+    return " ".join(words)
+
+
 def build_finished(app, exception):
     """
     Output YAML on the file system.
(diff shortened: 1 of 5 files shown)


docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,24 @@ def find_markdown_pages(app, outdir):
             })
 
 
+# Looks for cross references to look for in given content, except for
+# current_name.
+def find_cross_references(content, current_name, keyword_map):
```
tbpg, line 1266: "Looks for" and "find" do not sound like the content will actually be modified and returned with cross references replaced.

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,24 @@ def find_markdown_pages(app, outdir):
...
+    words = content.split(" ")
+    # Using counter to check if the entry is already a cross reference.
+    for i in range(0, len(words)):
+        for keyword in sorted(keyword_map.keys(), reverse=True):
```
tbpg, line 1270: Sort `keyword_map` once?

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,24 @@ def find_markdown_pages(app, outdir):
             })
 
 
+# Looks for cross references to look for in given content, except for
+# current_name.
+def find_cross_references(content, current_name, keyword_map):
+    words = content.split(" ")
+    # Using counter to check if the entry is already a cross reference.
+    for i in range(0, len(words)):
```
tbpg, line 1269: Use `enumerate`?

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,24 @@ def find_markdown_pages(app, outdir):
...
+        for keyword in sorted(keyword_map.keys(), reverse=True):
+            if keyword != current_name and keyword not in current_name and keyword in words[i]:
+                # If it is already processed as cross reference, skip over it.
+                if "<xref" in words[i-1]:
+                    continue
+                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
+                words[i] = words[i].replace(keyword, cross_reference)
```
tbpg, line 1276: We should avoid modifying the list we're iterating over.

docfx_yaml/extension.py
```diff
@@ -1446,6 +1465,100 @@ def convert_module_to_package_if_needed(obj):
...
+                #   SplitReadStreamResponse
+                # non-candidates:
+                #   (not from the same library)
+                #   google.cloud.aiplatform.AutoMLForecastingTrainingJob
+
+                current_name = obj["fullName"]
+                # Currently we only need to look in summary, syntax and
+                # attributes for cross references.
+                if obj.get("summary"):
+                    entry = "summary"
```
tbpg, line 1484: What is the `entry` variable for throughout?

docfx_yaml/extension.py
```diff
@@ -1446,6 +1465,100 @@ def convert_module_to_package_if_needed(obj):
...
+                    obj[entry] = find_cross_references(obj[entry], current_name, app.env.docfx_uid_names)
+
+                if obj.get("syntax"):
```
tbpg, line 1487: There is a lot of logic here. Would you mind moving all of these to another function (summary, syntax, and attributes)?

tests/test_helpers.py
```diff
@@ -57,5 +57,46 @@ def test_extract_keyword(self):
             keyword_got = extract_keyword(keyword_line)
 
 
+    def test_find_cross_references(self):
+        #Check that entries correctly turns into cross references.
```
tbpg, line 61: Nit:
```suggestion
        # Check that entries correctly turns into cross references.
```

tests/test_helpers.py
```diff
@@ -57,5 +57,46 @@ def test_extract_keyword(self):
             keyword_got = extract_keyword(keyword_line)
 
 
+    def test_find_cross_references(self):
```
tbpg, line 60: Can you pull some of these into a parameterized test? The test logic appears to be very similar for multiple.

tbpg (changes requested):

docfx_yaml/extension.py
dandhlee: Organized it into `search_cross_references` and renamed existing function to `convert_cross_references`.

docfx_yaml/extension.py
dandhlee: I wanted to highlight each of the sections by marking it with `entry = {section}`, but albeit being repetitive it makes more sense to use the raw string for better readability.

dandhlee (commented):
Thank you! Reading into parameterized tests, I'll add it in a bit.

dandhlee:
Addressed all comments, included unit test for the new function as well. Please take a look again!

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,117 @@ def find_markdown_pages(app, outdir):
...
+def convert_cross_references(content, current_name, sorted_map):
+    words = content.split(" ")
+    new_words = []
+    # Using counter to check if the entry is already a cross reference.
+    for index, word in enumerate(words):
+        cross_reference = ""
+        for keyword in sorted_map:
+            if keyword != current_name and keyword not in current_name and keyword in word:
+                # If it is already processed as cross reference, skip over it.
+                if "<xref" in words[index-1] or (new_words and f"<xref uid=\"{keyword}" in new_words[-1]):
```
tbpg, line 1275: When would this first `if` condition be true and the second be false?

* `"<xref" in words[index-1]`
* `... and f"<xref uid=\"{keyword}" in new_words[-1]`

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,117 @@ def find_markdown_pages(app, outdir):
             })
 
 
+# Finds and replaces occurrences which should be a cross reference in the given
+# content, except for the current name.
+def convert_cross_references(content, current_name, sorted_map):
```
tbpg, line 1266: Is there a more specific name for `sorted_map`? From the name, I don't know what it holds. Also, why does it need to be sorted? And `map` seems like not the right name for a list.

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,117 @@ def find_markdown_pages(app, outdir):
...
+    words = content.split(" ")
+    new_words = []
+    # Using counter to check if the entry is already a cross reference.
+    for index, word in enumerate(words):
+        cross_reference = ""
+        for keyword in sorted_map:
+            if keyword != current_name and keyword not in current_name and keyword in word:
```
tbpg, line 1273: Why is it `keyword in word` instead of equals?

tests/test_helpers.py
```diff
@@ -57,5 +59,104 @@ def test_extract_keyword(self):
...
+            "Response message for SplitReadStreamResponse."
+        ],
+        # Testing for cross references to convert within longer content.
+        [
+            "Response message for <xref uid=\"google.cloud.bigquery_storage_v1.types.SplitReadStreamResponse\">google.cloud.bigquery_storage_v1.types.SplitReadStreamResponse</xref>.",
+            "Response message for google.cloud.bigquery_storage_v1.types.SplitReadStreamResponse."
+        ],
+    ]
+    @parameterized.expand(cross_references_testdata)
+    def test_convert_cross_references(self, content_want, content):
```
tbpg, line 80: Nit: would you mind swapping the order of content and want? That way, when you're reading the test cases from top to bottom, you start with the input then see the expected output. On first read, I was confused why the `xref` was getting removed.

tbpg (changes requested):
A few more comments. Thank you!

docfx_yaml/extension.py
tbpg: Ahh. Re-reading... `words` vs `new_words`. So, yes, they are different. Ignore me. :smile:

docfx_yaml/extension.py
dandhlee: Ooooooops. Should've looked that more thoroughly :( Updated to reflect that it's a list.

I'm putting it into a sorted list so that the shorter version of the match does not take the cross reference conversion before the appropriate match. For example, without the sorted list in the following content:
```
var_type: google.cloud.bigquery_storage_v1.types.ReadSession
```
we expect the following:
```
var_type: <xref uid="google.cloud.bigquery_storage_v1.types.ReadSession">google.cloud.bigquery_storage_v1.types.ReadSession</xref>
```
however, we could and will encounter the following:
```
var_type: <xref uid="google.cloud.bigquery_storage_v1.types">google.cloud.bigquery_storage_v1.types</xref>.ReadSession
```

Looking back, this doesn't sound intuitive to me either but since we're looking for `keyword in word` and can't do an exact match and also know if there is more we should consider than the match we have (for example, when google.cloud.bigquery_storage_v1.types is a match how do we know there's more?), this is the best I can think of for now. Happy to explore better alternatives!

docfx_yaml/extension.py
dandhlee: Because we break by space, we could encounter situations like `` `google.cloud.bigquery_storage_v1.types` `` and other similar forms where the cross reference exists within special characters or as part of some form of one line syntax.

docfx_yaml/extension.py
dandhlee: 👍

tests/test_helpers.py
dandhlee: Agreed. Done.

dandhlee (commented):
Done. Please see the comments below.

docfx_yaml/extension.py
```diff
@@ -1260,6 +1261,117 @@ def find_markdown_pages(app, outdir):
             })
 
 
+# Finds and replaces occurrences which should be a cross reference in the given
+# content, except for the current name.
+def convert_cross_references(content, current_name, list_of_entry_names):
```
tbpg, line 1266: Total nit: I think just `entry_names` is enough? Sounds like it would be a list.

tbpg (approved):

docfx_yaml/extension.py
busunkim96: If you feel it's unclear you could alternatively add type hints here. :)

dandhlee: Adding type hints sounds like a good idea! I'll keep this in mind when refactoring the code as well.

PR #144, opened 2021-11-14
feat: format signature using black

dandhlee:
Figured out that `black` could be called within Python using `format_str` function. Using this to better format function signatures for our docs.

Requires a bit of pre-processing the data to look like actual code, otherwise black will not be able to process the given code as it's meant to run in files. I also plan on using this to further format code for other types of code so I've created a standalone function. Will revisit with more types of code to format, for now it will only be formatting signatures.

Fixes b/199327840

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -23,6 +23,7 @@ import inspect
 import re
 import copy
 import shutil
+import black
 from pathlib import Path
 from functools import partial
 from itertools import zip_longest
@@ -712,11 +713,12 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
                     if arg in type_map:
                         arg_map['var_type'] = type_map[arg]
                         args.append(arg_map)
+
             if argspec.varargs:
                 args.append({'id': argspec.varargs})
             if argspec.varkw:
                 args.append({'id': argspec.varkw})
-            # Try to add default values. Currently does not work if there is * argument present.
+
             if argspec.defaults:
                 # Attempt to add default values to arguments.
                 try:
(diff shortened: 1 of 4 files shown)


dandhlee:
visual output comparison:

before formatting:
```python
batch_predict(*, gcs_source: Optional[Union[str, Sequence[str]]] = None, instances_format: str = "jsonl", gcs_destination_prefix: Optional[str] = None, predictions_format: str = "jsonl", model_parameters: Optional[Dict] = None, machine_type: Optional[str] = None, accelerator_type: Optional[str] = None, explanation_parameters: Optional[google.cloud.aiplatform_v1.types.explanation.ExplanationParameters] = None, labels: Optional[Dict[str, str]] = None, sync: bool = True,)
```

after formatting:
```python
batch_predict(
    *,
    gcs_source: Optional[Union[str, Sequence[str]]] = None,
    instances_format: str = "jsonl",
    gcs_destination_prefix: Optional[str] = None,
    predictions_format: str = "jsonl",
    model_parameters: Optional[Dict] = None,
    machine_type: Optional[str] = None,
    accelerator_type: Optional[str] = None,
    explanation_parameters: Optional[
        google.cloud.aiplatform_v1.types.explanation.ExplanationParameters
    ] = None,
    labels: Optional[Dict[str, str]] = None,
    sync: bool = True,
)
```

docfx_yaml/extension.py
```diff
@@ -959,10 +961,19 @@ def process_docstring(app, _type, name, obj, options, lines):
     app.env.docfx_info_uid_types[datam['uid']] = _type
 
 
+# Uses black.format_str() to reformat code as if running black/linter
+# for better presnetation.
+def format_code(code):
+    # Signature code comes in raw text without formatting, to run black it
+    # requires the code to look like actual function declaration in code.
+    # Returns the original formatted code without the added bits.
+    return black.format_str("def " + code + ": pass", mode=black.FileMode())[4:-11]
```
tbpg, line 970: Nice!

tbpg (approved):

docfx_yaml/extension.py
plamut: I like this too, but it might be fragile. How is `code` generated?

The problem seems to be with representations of default values that are not `eval()`-uable, such as the default timeout object.

An [error](https://source.cloud.google.com/results/invocations/38ad16c2-d98a-49b4-88a7-c90163dcf7b1/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fdocs%2Fdocs-presubmit/log) from a bigquery build:
```
Cannot parse: 1:58: def cancel_job(job_id, project=None, location=None, retry=<google.api_core.retry.Retry object>, timeout=None): pass
```

dandhlee: The `code` is found from `signature` variable at this point: https://github.com/googleapis/sphinx-docfx-yaml/blob/4e1d03fe1bc78a67656a3f71fb93981e8fd3e255/docfx_yaml/extension.py#L984

Which is passed from Sphinx in https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/ext/autodoc/__init__.py#L498. At this point it seems like we can't resolve the references so I've filed #151 to look into this.

PR #170, opened 2022-01-21
feat: add syntax highlighting support for Markdown pages

dandhlee:
Markdown pages that are in `rst` format files come without any syntax highlighting support as they're omitted by the Sphinx markdown plugin. For all Markdown pages this is expected for none of the code blocks (whether specified to include a language or not) to have language specified to help with syntax highlighting, hence this feature will likely work well. In cases where there might be few that slip through the crack or are malformed, I've included an error check to ensure that we only process ones that seem valid.

Once `python` tag is added, actual syntax-highlight is taken care of by the [code](https://github.com/googleapis/doc-pipeline/blob/main/docpipeline/prepare.py#L16-L27) in `doc-pipeline`. 

If this is added to a code block that's not Python code, `prettyprint` attempts to find a language based on the code provided and `lang-python` is just a supplement to help clarify that it might be Python code but will not add Python syntax highlight to Java code per se.

Fixes b/213152730.

- [x] Tests pass


docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,45 @@ def extract_header_from_markdown(mdfile_iterator):
...
+# For a given markdown file, adds syntax highlighting to code blocks.
+def highlight_md_codeblocks(mdfile):
+    find_string = '```'
+    find_string_nl = '```\n'
+    replace_string = '```python'
+    new_lines = []
+
+    with open(mdfile) as mdfile_iterator:
+        file_content = mdfile_iterator.read()
+        # If there isn't even number of code block annotations do not syntax
```
tbpg, line 1280:
```suggestion
        # If there is an odd number of code block annotations, do not syntax
```

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,45 @@ def extract_header_from_markdown(mdfile_iterator):
     return mdfile_name
 
 
+# For a given markdown file, adds syntax highlighting to code blocks.
+def highlight_md_codeblocks(mdfile):
+    find_string = '```'
+    find_string_nl = '```\n'
+    replace_string = '```python'
```
tbpg, line 1275: Consider `fence`, `fence_with_nl`, and `fence_with_python`, or something like that. It's a little unclear to me while reading what these names refer to. Could just be me.

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,45 @@ def extract_header_from_markdown(mdfile_iterator):
...
+    new_lines = []
+
+    with open(mdfile) as mdfile_iterator:
+        file_content = mdfile_iterator.read()
+        # If there isn't even number of code block annotations do not syntax
+        # highlight.
+        if file_content.count(find_string_nl) % 2 != 0:
```
tbpg, line 1282: What happens if a code block actually has a language indicator? For example, this example has valid code fencing, but this check would return an odd number.

````
```python
oops
```
````

File: tests/markdown_mixed_post.md [hunk 4]
```diff
@@ -0,0 +1,7 @@
+```
+File with missing codeblock
+```
+
+```
+with no closing bracket
```
tbpg, line 6: Consider renaming the test files following the pattern `foo.md` and `foo_want.md`. That way, we're consistent between filenames and the tests themselves.

tests/markdown_code_obj.md
```diff
@@ -0,0 +1,9 @@
+```
+test markdown file for
+highlighing markdown codeblocks
+```
+
+```
+all code blocks
+should be highlighted
+```
```
tbpg, line 9: Please add a test including a code block with a language indicator.

tbpg (changes requested):

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,45 @@ def extract_header_from_markdown(mdfile_iterator):
...
+            print(f'{mdfile_iterator.name} contains mixed or wrong format of code blocks. Skipping syntax highlighting.')
+            return
+        # Retrieve code block positions to replace
+        codeblocks = [[m.start(), m.end()] for m in re.finditer(
+                                                      find_string,
+                                                      file_content)]
+
+        # This is equivalent to grabbing every odd index items.
```
busunkim96, line 1290:
```suggestion
        # This is equivalent to grabbing every odd index item.
```

tests/markdown_code_obj.md
busunkim96: +1, I've seen both `py` and `python` language indicators in repos.

docfx_yaml/extension.py
dandhlee: Thank you! Didn't know they were called fence 😅

docfx_yaml/extension.py
dandhlee: Applied suggestion.

docfx_yaml/extension.py
dandhlee: For what I've seen throughout the libraries this never was the case, nonetheless I should ensure that this case is handled. At the moment we'd return without syntax highlighting applied in the file but this is also not going to work if there's exactly even number of language indicators which would cause chaos.

At the moment the easiest solution should be to check for language indicators' existence, and omit those pair of entries. I'll modify this code a bit to still check that there's an even number of fences.

docfx_yaml/extension.py
dandhlee: Applied suggestion.

tests/markdown_mixed_post.md
dandhlee: Done.

tests/markdown_code_obj.md
dandhlee: Done, added support for language indicators and unit test for it.

dandhlee (commented):
Thank you both! Please take a look again.

docfx_yaml/extension.py
```diff
@@ -1294,6 +1343,7 @@ def find_markdown_pages(app, outdir):
     # For each file, if it is a markdown file move to the top level pages.
     for mdfile in markdown_dir.iterdir():
         if mdfile.is_file() and mdfile.name.lower() not in files_to_ignore:
+            highlight_md_codeblocks(f"{markdown_dir}/{mdfile.name}")
```
tbpg, line 1346: Why the string/name conversion?

tests/markdown_mixed_highlight.md
```diff
@@ -0,0 +1,15 @@
+```python
+These code block should not be highlighted.
```
tbpg, line 2:
```suggestion
These code blocks should not be highlighted.
```

tests/markdown_mixed_highlight.md
```diff
@@ -0,0 +1,15 @@
...
+```
+
+```py
+As these comes with a language indicator.
```
tbpg, line 6:
```suggestion
As these come with a language indicator.
```

tests/test_helpers.py
```diff
@@ -190,5 +194,41 @@ def test_extract_product_name(self):
 
         self.assertEqual(short_name_want, short_product_name)
 
+
+    # Filenames to test markdown syntax highlight with.
+    test_markdown_filenames = [
+        [
+            "tests/markdown_syntax_highlight.md",
+            "tests/markdown_syntax_highlight_got.md",
```
tbpg, line 202: We should either use a tmp file, clean these up every time (not after the assert), or include them in `.gitignore`.

tbpg (changes requested):

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,55 @@ def extract_header_from_markdown(mdfile_iterator):
...
+        for start, end in codeblocks:
+            newline_index = file_content.find('\n', start)
+            # There should be exactly 3 characters between the start of the
+            # fence and the newline, otherwise there probably is a language
+            # indicator.
+            if newline_index - start == 3:
+                blocks_without_indicators.append([start, end])
+
+        # Stitch content that does not need to be parsed, and replace with
+        # `replace_string` for parsed portions.
```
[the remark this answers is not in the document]
dandhlee: Updated reference to old variable name.

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,55 @@ def extract_header_from_markdown(mdfile_iterator):
...
+                                                      file_content)]
+
+        # This is equivalent to grabbing every odd index item.
+        codeblocks = codeblocks[::2]
+        # Used to store code blocks that comes without language indicators.
+        blocks_without_indicators = []
+
+        # Check if the fence comes with a language indicator. If so, skip this.
+        for start, end in codeblocks:
+            newline_index = file_content.find('\n', start)
```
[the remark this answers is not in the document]
dandhlee: Checking one character after the end of the fence makes much more sense. Updated!

docfx_yaml/extension.py
```diff
@@ -1268,6 +1268,55 @@ def extract_header_from_markdown(mdfile_iterator):
...
+def highlight_md_codeblocks(mdfile):
+    fence = '```'
+    fence_with_python = '```python'
+    new_lines = []
+
+    with open(mdfile) as mdfile_iterator:
+        file_content = mdfile_iterator.read()
+        # If there is an odd number of code block annotations, do not syntax
+        # highlight.
+        if file_content.count(fence) % 2 != 0:
```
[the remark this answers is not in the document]
dandhlee: The background of this issue is that the Markdown plugin that's generating all of the markdown pages, is simply not adding the language indicators to the code fences. If there are four backticks in the code fence, it will not be generating from the Markdown plugin but directly inserted into the file which we can simply fix it in the source (or ask to include it with a language indicator).

I'd prefer this be solved at the Markdown plugin level as well :( Happy to file an issue on this to track for future improvement.

dandhlee (commented):

tests/test_helpers.py
```diff
@@ -190,5 +193,41 @@ def test_extract_product_name(self):
...
+            "tests/markdown_mixed_highlight.md",
+            "tests/markdown_mixed_highlight_want.md"
+        ],
+    ]
+    @parameterized.expand(test_markdown_filenames)
+    def test_highlight_md_codeblocks(self, base_filename, want_filename):
+        # Test to ensure codeblocks in markdown files are correctly highlighted.
+
+        # Copy the base file we'll need to test.
+        test_file = tempfile.NamedTemporaryFile(mode='r+', delete=False)
```
tbpg, line 217: Use as a context manager (`with ...`) to avoid having to manually close? Why not auto-delete? Debugging if it goes wrong?

dandhlee: The file needed to be kept within the function being tested, so I didn't think it'd work to use it as a context manager. Seems to work, updated to use.

tbpg (approved):

PR #175, opened 2022-01-28
test: disable 3.6 test

### LINKED ISSUE #174: Disable 3.6 test
Python 3.6 is now end of life, and 3.10 has been added for a while.

dandhlee:
Updating the unit test with appropriate Python versions.

Fixes #174.

- [x] Tests pass


busunkim96 (approved):
It would be good to also add a [`python_requires`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires) to the `setup.py`.

PR #181, opened 2022-02-12
fix: retrieve constructors

dandhlee:
Because of a Sphinx configuration set up on the Client libraries, Sphinx will try and combine the constructor and the Class definition together. However, from the Plugin side this is encountered as two separate docstrings with same name and `_type == CLASS`. Since we tried avoiding duplicate entries, this ended up getting rid of the constructors.

Retrieved constructors by identifying them (on their second encounter) as a method. The original class definition will retain much information about it as it can, with minimal overlap between the two. There can be an overlap for the constructor syntax showing both on the class header definition and the constructor method definition, I will ask for feedback on whether this should be okay or not.

As well, removing arguments with empty `var_type` or `description`. Missing either two likely results from a typo or from incomplete documentation which should be brought up with the respective client library owners. 

Fixes b/208869740, b/208866895.

- [ ] Tests pass: tests will be part of the goldens for work to be done in #44.

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -717,7 +717,7 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
     # Check how many arguments are present in the function.
     arg_count = 0
     try:
-        if _type in [METHOD, FUNCTION]:
+        if _type in [METHOD, FUNCTION, CLASS]:
             argspec = inspect.getfullargspec(obj) # noqa
             type_map = {}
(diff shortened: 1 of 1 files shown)


busunkim96 (commented):
> As well, removing arguments with empty var_type or description. Missing either two likely results from a typo or from incomplete documentation which should be brought up with the respective client library owners.

Could you share an example of a docstring where this would happen?

busunkim96 (approved):

PR #193, opened 2022-03-18
fix: format code snippets properly

dandhlee:
Code snippets that are separated by new blocks were not treated with the same `tab_space`, resulting in inconsistent formatting between the blocks. To make this consistent, the parser will now collect all of the code snippet first, then process all of them at once with a uniform `tab_space` value that was calculated when the code block is encountered at first.

Modified the unit test to grab the `tab_space` like how the parser will compute it. Also added an additional unit test for testing code snippet with blocks and refactored the existing unit test to use parameterized style instead.

Fixes b/219067629.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -388,11 +388,8 @@ def extract_keyword(line):
 
 # Given lines of code, indent to left by 1 block, based on
 # amount of trailing white space of first line as 1 block.
-def indent_code_left(lines):
+def indent_code_left(lines, tab_space):
     parts = lines.split("\n")
-    # Count how much leading whitespaces there are based on first line.
-    # lstrip(" ") removes all trailing whitespace from the string.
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -388,11 +388,8 @@ def extract_keyword(line):
 
 # Given lines of code, indent to left by 1 block, based on
 # amount of trailing white space of first line as 1 block.
-def indent_code_left(lines):
+def indent_code_left(lines, tab_space):
```
tbpg, line 391: Can you add a doc comment saying what these args are?

tbpg (approved):
One minor doc comment.

docfx_yaml/extension.py
dandhlee: Done.

dandhlee:
Thank you! Filed #194 to keep track of converting rest of comments to the docstring format.

PR #200, opened 2022-04-04
fix: support parsing summary docstring that is not well formed

dandhlee:
Currently, there are `.. code-block` docstrings that could appear with extra newlines if the summary before the docstring is longer than one line (80 characters). This causes the parser to not correctly find the docstring element. The parser however should be able to correctly locate the docstring despite the summary not being properly formatted, so the users do not see extra items in the documentation.

Fixing this by stripping any existence of newlines that come before the docstring; any other characters will still not be tolerated. 

Modified the unit test for a code block that contains an extra newline; without the fix in this code, the unit test will fail.

Fixes b/219067629.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -469,9 +469,13 @@ def _parse_docstring_summary(summary):
                 continue
 
         # Parse keywords if found.
-        if part.startswith(".."):
+        # lstrip is added to parse code blocks that are not formatted well.
+        # TODO: remove the lstrip if
+        # https://github.com/googleapis/gapic-generator-python/issues/1244 is
+        # resolved.
+        if part.lstrip('\n').startswith('..'):
             try:
-                keyword = extract_keyword(part)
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -469,9 +469,13 @@ def _parse_docstring_summary(summary):
                 continue
 
         # Parse keywords if found.
-        if part.startswith(".."):
+        # lstrip is added to parse code blocks that are not formatted well.
+        # TODO: remove the lstrip if
+        # https://github.com/googleapis/gapic-generator-python/issues/1244 is
+        # resolved.
```
tbpg, line 475: Do we want to do this if older versions will always have the leading newlines?

tbpg (approved):

docfx_yaml/extension.py
dandhlee: Nay, we should keep it for it to be backwards looking. Removed the to-do comment.

PR #203, opened 2022-04-12
fix: add markdown header if it is missing

### LINKED ISSUE #135: Render and format markdown pages within the plugin
Googlers see b/206694500.

After the markdown pages are processed, we should make them look more pretty, and also include a way to filter some of these pages.

Few things to consider right now:

1. `multiprocessing.md` files currently show with a really large header, since it's missing a header. For files missing a header, we should add in a header as well.
~~2. Sometimes the headers are too long, and most header items get extracted (H1~H3) and gets shown on the right-nav. We should find a way to shorten and make it look more pretty when possible.~~

dandhlee:
A markdown file may come without a title. Ideally, this could get fixed from the source, but the plugin should try and render content that is still readable rather than not making the rest of the content unreadable and giving the user poorer experience.

This means that if a title exists but do not conform to the Markdown's style, there would be duplicated content. However, I think that is much better rather than a completely non-readable title and body content. For example, I'd prefer

```
<h1> Markdown_title </h1>
 #Markdown_title

This is the body content of the markdown page with a malformed title.
```

rather than
```
<h1> #Markdown_title This is the body content of the markdown page with a malformed title. </h1>
```

If the owner of the documentation page does not like seeing the former, they can simply update the source to have a proper title and the plugin will use the correct title as the header.

Adding unit test and updating existing tests as `extract_header_from_markdown` now returns an empty string instead of a proposed title. 

Fixes #135 and b/208866831.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1287,8 +1287,8 @@ def parse_markdown_header(header_line, prev_line):
 
 
 # For a given markdown file, extract its header line.
+# Returns empty string if a Markdown heading level 1 does not exist.
 def extract_header_from_markdown(mdfile_iterator):
-    mdfile_name = mdfile_iterator.name.split("/")[-1].split(".")[0].capitalize()
     prev_line = ""
 
     for header_line in mdfile_iterator:
@@ -1301,8 +1301,7 @@ def extract_header_from_markdown(mdfile_iterator):
 
         prev_line = header_line
 
-    print(f"Could not find a title for {mdfile_iterator.name}. Using {mdfile_name} as the title instead.")
-    return mdfile_name
+    return ""
 
 
 # For a given markdown file, adds syntax highlighting to code blocks.
@@ -1351,6 +1350,20 @@ def highlight_md_codeblocks(mdfile):
         mdfile_iterator.write(new_content)
 
 
+def prepend_markdown_title(filename, mdfile_iterator):
+    """Prepends the name as a Markdown title.
+
+    Args:
+        filename: the name of the markdown file to prepend.
+        mdfile_iterator: iterator to the markdown file that is both readable
+        and writable.
+    """
+    file_content = f'# {filename}\n\n' + mdfile_iterator.read()
+    # Reset file position to the beginning to write
+    mdfile_iterator.seek(0)
+    mdfile_iterator.write(file_content)
+
+
 # Given generated markdown files, incorporate them into the docfx_yaml output.
 # The markdown file metadata will be added to top level of the TOC.
 def find_markdown_pages(app, outdir):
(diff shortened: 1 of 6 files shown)


docfx_yaml/extension.py
```diff
@@ -1351,6 +1350,20 @@ def highlight_md_codeblocks(mdfile):
         mdfile_iterator.write(new_content)
 
 
+def prepend_markdown_header(filename, mdfile_iterator):
+    """Prepends the name as a Markdown header.
```
tbpg, line 1354:
```suggestion
    """Prepends the filename as a Markdown header.
```

tests/test_helpers.py
```diff
@@ -260,5 +261,40 @@ def test_highlight_md_codeblocks(self, base_filename, want_filename):
...
+            "tests/markdown_example_h2_want.md"
+        ],
+        [
+            "tests/markdown_example_alternate_bad.md",
+            "tests/markdown_example_alternate_bad_want.md"
+        ],
+    ]
+    @parameterized.expand(test_markdown_filenames)
+    def test_prepend_markdown_header(self, base_filename, want_filename):
+        # Test to ensure markdown titles are correctly prepended.
```
tbpg, line 281:
```
"""Ensure..."""
```

tests/test_unit.py
```diff
@@ -713,24 +713,20 @@ def test_extract_header_from_markdown_alternate_header(self):
 
     def test_extract_header_from_markdown_bad_headers(self):
         # Check that the filename is used as header if no valid header is found.
```
tbpg, line 715: Update.

tests/test_unit.py
```diff
@@ -713,24 +713,20 @@ def test_extract_header_from_markdown_alternate_header(self):
 
     def test_extract_header_from_markdown_bad_headers(self):
```
tbpg, line 714: Seems like this one should also be parameterized.

tbpg (changes requested):

tests/test_helpers.py
dandhlee: Done.

tests/test_unit.py
dandhlee: Turned relevant tests to be parameterized.

tests/test_unit.py
dandhlee: Done.

dandhlee (commented):

tests/test_unit.py
```diff
@@ -667,67 +668,69 @@ def test_parse_markdown_header_alternate(self):
...
-            header_line_got = extract_header_from_markdown(mdfile)
-
-        self.assertEqual(header_line_want, header_line_got)
-
-        # Check that there must be a line before the h1 header breaker.
-        with open('tests/markdown_example_alternate_bad.md', 'r') as mdfile:
+        with open(markdown_filename, 'r') as mdfile:
             header_line_got = extract_header_from_markdown(mdfile)
 
         self.assertEqual(header_line_want, header_line_got)
```
tbpg, line 736: assertEmpty

docfx_yaml/extension.py
```diff
@@ -1351,6 +1350,20 @@ def highlight_md_codeblocks(mdfile):
         mdfile_iterator.write(new_content)
 
 
+def prepend_markdown_header(filename, mdfile_iterator):
```
tbpg, line 1353: I thought I commented with this before, but I don't see it. Consider using type hints instead of naming the arg `..._iterator`.

tests/test_unit.py
dandhlee: I thought assertEmpty exists but assertTrue/assertFalse is used instead apparently 😮

tbpg (approved):

docfx_yaml/extension.py
```diff
@@ -1286,12 +1286,18 @@ def parse_markdown_header(header_line, prev_line):
     return ""
 
 
-# For a given markdown file, extract its header line.
-def extract_header_from_markdown(mdfile_iterator):
-    mdfile_name = mdfile_iterator.name.split("/")[-1].split(".")[0].capitalize()
+def extract_header_from_markdown(mdfile: Iterable) -> str:
```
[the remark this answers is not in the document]
dandhlee: Done. Wasn't sure what type to use for file content :S

docfx_yaml/extension.py
```diff
@@ -1351,6 +1356,20 @@ def highlight_md_codeblocks(mdfile):
         mdfile_iterator.write(new_content)
 
 
+def prepend_markdown_header(filename: str, mdfile: Iterable):
```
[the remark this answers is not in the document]

Issue #208, opened 2022-05-17
Remove hardcoded reference to grpc-google-iam


dandhlee:
For references to `google.iam.v1.iam_policy_pb2` methods, there is no appropriate documentation available at the moment. After the IAM libraries properly get merged and there is a documentation I could point to, remove the hard coded reference.

PR #209, opened 2022-05-17
fix: add hardcoded IAM references temporarily

dandhlee:
At the moment, there are 4 IAM libraries and it is very unclear to users which IAM library is referenced in the documentation. To make this worse, some of them don't have documentation that is generated to point to. The IAM team is reviewing combining all the libraries under `python-iam`, but until then it will remain unclear.

To temporarily address this, pointing the IAM references to the GitHub repository where the source code is located. This hard coded reference list can also be used for other common libraries that the documentation is missing references to, until we can figure out Intersphinx and hosting that on c.g.c.

Filed https://github.com/googleapis/sphinx-docfx-yaml/issues/208 to track removing the hard coded references in the future.

Fixes b/232773389.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1418,21 +1418,21 @@ def find_markdown_pages(app, outdir):
             })
 
 def resolve_cross_reference(
+    current_word: str,
     words: List[str],
     index: int,
-    current_word: str,
     keyword: str,
-    visited_words: List[str],
+    processed_words: List[str],
     hard_coded_references: Dict[str, str] = None
 ) -> Optional[str]:
-    """Given a word, check if it contains a reference keyword we should convert.
+    """Given `current_word`, returns the resolved cross reference for `keyword` if found.
 
     Args:
+        current_word: current word being looked at
         words: list of words used to check and compare content before and after `current_word`
         index: index position of `current_word` within words
-        current_word: current word being looked at
         keyword: reference keyword to check against
-        visited_words: list of words containing words that's been processed so far
+        processed_words: list of words containing words that's been processed so far
         hard_coded_references: Optional list containing a list of hard coded reference
 
     Returns:
@@ -1441,14 +1441,21 @@ def resolve_cross_reference(
     """
     if keyword in current_word:
         if hard_coded_references and keyword in hard_coded_references:
+            # If the cross reference has been processed already, "<a" will
+            # appear as the previous word. Also check against the words that
+            # have been converted. For hard coded references, we use
+            # "<a href" links.
             if "<a" not in words[index-1] and \
-                not (visited_words and f"<a href=\"{hard_coded_references[keyword]}" in visited_words[-1]):
+                not (processed_words and f"<a href=\"{hard_coded_references[keyword]}" in processed_words[-1]):
                 return f"<a href=\"{hard_coded_references[keyword]}\">{keyword}</a>"
 
         else:
-            # If it is already processed as cross reference, skip over it.
+            # Check to see if the keyword has always been identified as a cross
+            # reference. If so, the original text will contain "<xref" as the
+            # previous word, i.e. "<xref uid="storage">storage</xref>". Check
+            # against the words that have been converted so far as well.
             if "<xref" not in words[index-1] and \
-                not (visited_words and f"<xref uid=\"{keyword}" in visited_words[-1]):
+                not (processed_words and f"<xref uid=\"{keyword}" in processed_words[-1]):
                 return f"<xref uid=\"{keyword}\">{keyword}</xref>"
 
     return None
@@ -1469,13 +1476,12 @@ def convert_cross_references(content: str, current_object_name: str, entry_names
 
     Returns:
         content that has been modified with proper cross references if found.
-
     """
     words = content.split(" ")
 
     # Contains a list of words that is not a valid reference or converted
     # references.
-    visited_words = []
+    processed_words = []
 
     # TODO(https://github.com/googleapis/sphinx-docfx-yaml/issues/208):
     # remove this in the future.
@@ -1492,23 +1498,28 @@ def convert_cross_references(content: str, current_object_name: str, entry_names
     for index, word in enumerate(words):
         cross_reference = None
         for keyword in entry_names:
-            # Do not convert references to itself.
+            # Do not convert references to itself or containing partial
+            # references. This could result in `storage.types.ReadSession` being
+            # prematurely converted to
+            # `<xref uid="storage.types">storage.types</xref>ReadSession`
+            # instead of
+            # `<xref uid="storage.types.ReadSession">storage.types.ReadSession</xref>`
             if keyword in current_object_name:
                 continue
 
             cross_reference = resolve_cross_reference(
-                words, index, word, keyword, visited_words, hard_coded_references
+                word, words, index, keyword, processed_words, hard_coded_references
             )
             if cross_reference:
-                visited_words.append(word.replace(keyword, cross_reference))
+                processed_words.append(word.replace(keyword, cross_reference))
                 print(f"Converted {keyword} into cross reference in: \n{content}")
                 break
 
         # If cross reference has not been found, add current unchanged content.
         if not cross_reference:
-            visited_words.append(word)
+            processed_words.append(word)
 
-    return " ".join(visited_words)
+    return " ".join(processed_words)
 
 
 # Used to look for cross references in the obj's data where applicable.
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1423,11 +1423,32 @@ def find_markdown_pages(app, outdir):
 def convert_cross_references(content: str, current_name: str, entry_names: List[str]):
     words = content.split(" ")
     new_words = []
+
+    # TODO(https://github.com/googleapis/sphinx-docfx-yaml/issues/208):
+    # remove this in the future.
+    iam_policy_link = "http://github.com/googleapis/python-grpc-google-iam-v1/blob/main/google/iam/v1/iam_policy_pb2_grpc.py"
+    hard_coded_references = {
+        "google.iam.v1.iam_policy_pb2.SetIamPolicyRequest": iam_policy_link + "#L103-L109",
+        "google.iam.v1.iam_policy_pb2.GetIamPolicyRequest": iam_policy_link + "#L111-L118",
+        "google.iam.v1.iam_policy_pb2.TestIamPermissionsRequest": iam_policy_link + "#L120-L131",
+        "google.iam.v1.iam_policy_pb2.TestIamPermissionsResponse": iam_policy_link + "#L120-L131"
+    }
+    entry_names.extend(hard_coded_references.keys())
+
     # Using counter to check if the entry is already a cross reference.
     for index, word in enumerate(words):
         cross_reference = ""
         for keyword in entry_names:
             if keyword != current_name and keyword not in current_name and keyword in word:
+                if keyword in hard_coded_references:
+                    if "<a" in words[index-1] or \
+                        (new_words and f"<a href=\"{hard_coded_references[keyword]}" in new_words[-1]):
+                        continue
+                    cross_reference = f"<a href=\"{hard_coded_references[keyword]}\">{keyword}</a>"
+                    new_words.append(word.replace(keyword, cross_reference))
+                    print(f"Converted {keyword} into cross reference in: \n{content}")
+                    continue
+
                 # If it is already processed as cross reference, skip over it.
                 if "<xref" in words[index-1] or (new_words and f"<xref uid=\"{keyword}" in new_words[-1]):
                     continue
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1417,11 +1417,12 @@ def find_markdown_pages(app, outdir):
                 'href': mdfile.name.lower(),
             })
 
-def resolve_cross_reference(
+def find_uid_to_convert(
     current_word: str,
     words: List[str],
     index: int,
-    uid: str,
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -1423,11 +1423,32 @@ def find_markdown_pages(app, outdir):
 def convert_cross_references(content: str, current_name: str, entry_names: List[str]):
     words = content.split(" ")
     new_words = []
+
+    # TODO(https://github.com/googleapis/sphinx-docfx-yaml/issues/208):
+    # remove this in the future.
+    iam_policy_link = "http://github.com/googleapis/python-grpc-google-iam-v1/blob/main/google/iam/v1/iam_policy_pb2_grpc.py"
+    hard_coded_references = {
+        "google.iam.v1.iam_policy_pb2.SetIamPolicyRequest": iam_policy_link + "#L103-L109",
```
tbpg, line 1431: We should probably do some sort of query (e.g. https://github.com/googleapis/python-grpc-google-iam-v1/search?q=setiampolicy+path%3Agoogle%2Fiam%2Fv1%2F+filename%3Aiam_policy_pb2_grpc.py) or use a permalink (e.g. https://github.com/googleapis/python-grpc-google-iam-v1/blob/8e73b45993f030f521c0169b380d0fbafe66630b/google/iam/v1/iam_policy_pb2_grpc.py#L103).

docfx_yaml/extension.py
```diff
@@ -1423,11 +1423,32 @@ def find_markdown_pages(app, outdir):
...
+    }
+    entry_names.extend(hard_coded_references.keys())
+
     # Using counter to check if the entry is already a cross reference.
     for index, word in enumerate(words):
         cross_reference = ""
         for keyword in entry_names:
             if keyword != current_name and keyword not in current_name and keyword in word:
+                if keyword in hard_coded_references:
+                    if "<a" in words[index-1] or \
```
tbpg, line 1444: The indentation here is pretty gnarly (`for` -> `for` -> `if` -> `if` -> `if` with a couple `continue`s). Can we simplify this? Perhaps with a helper function?

docfx_yaml/extension.py
```diff
@@ -1423,11 +1423,32 @@ def find_markdown_pages(app, outdir):
...
+        "google.iam.v1.iam_policy_pb2.TestIamPermissionsResponse": iam_policy_link + "#L120-L131"
+    }
+    entry_names.extend(hard_coded_references.keys())
+
     # Using counter to check if the entry is already a cross reference.
     for index, word in enumerate(words):
         cross_reference = ""
         for keyword in entry_names:
             if keyword != current_name and keyword not in current_name and keyword in word:
+                if keyword in hard_coded_references:
```
tbpg, line 1443:
```suggestion
                if keyword, reference in hard_coded_references.items():
```

tbpg (changes requested):

docfx_yaml/extension.py
dandhlee: Done, updated to use a helper function.

dandhlee:

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+        if keyword in hard_coded_references:
+            if "<a" not in words[index-1] and \
+                not (converted_words and f"<a href=\"{hard_coded_references[keyword]}" in converted_words[-1]):
+                cross_reference = f"<a href=\"{hard_coded_references[keyword]}\">{keyword}</a>"
+
+        else:
+            # If it is already processed as cross reference, skip over it.
+            if "<xref" not in words[index-1] and \
+                not (converted_words and f"<xref uid=\"{keyword}" in converted_words[-1]):
+                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
```
tbpg, line 1456: Return.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+        current_name: the name of the current Python object being looked at
+        entry_names: list of keyword references to look for
+
+    Returns:
+        content that has been modified with proper cross references if found.
+
+    """
     words = content.split(" ")
-    new_words = []
+    converted_words = []
```
tbpg, line 1478: Why a list instead of a dictionary from word to converted word?

tbpg (changes requested):

docfx_yaml/extension.py
dandhlee: I suppose the name is still throwing off. As we traverse the original list of words, we still need to check against the converted list of words and order matters. At the end, I use this to simply convert to a string using a join. Will update the name so it's not misleading.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+
+    return cross_reference
+
 
-# Finds and replaces occurrences which should be a cross reference in the given
-# content, except for the current name.
 def convert_cross_references(content: str, current_name: str, entry_names: List[str]):
+    """Finds and replaces references that should be a cross reference in given content.
+
+    This should not convert any references that contain its name, i.e. if we're processing
```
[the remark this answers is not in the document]
dandhlee: Yes. Updated comment to clarify this.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+def check_existing_references(
+    current_name: str,
+    words: List[str],
+    index: int,
+    word: str,
+    keyword: str,
+    converted_words: List[str],
+    hard_coded_references: Dict[str, str] = None
+):
+    """Given a word, check if it contains a reference keyword we should convert.
```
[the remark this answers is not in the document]
dandhlee: I felt icky mixing between `None` and an empty string, however I think in this case it makes sense that we indeed return `None` instead of an empty string if a cross reference has not been found. Updating with suggestions.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
 
+def check_existing_references(
+    current_name: str,
+    words: List[str],
+    index: int,
+    word: str,
+    keyword: str,
+    converted_words: List[str],
+    hard_coded_references: Dict[str, str] = None
+):
```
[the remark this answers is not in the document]
dandhlee: Applied suggestion in separate commit.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
                 'href': mdfile.name.lower(),
             })
 
+def check_existing_references(
+    current_name: str,
+    words: List[str],
+    index: int,
+    word: str,
```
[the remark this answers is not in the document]
dandhlee: Updated `word` -> `current_word` and `current_name` -> `current_object_name`. However, checking for it in this function seemed useless so I removed it from this function and moved it up, happy to take any other variable name suggestions still.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+                not (converted_words and f"<a href=\"{hard_coded_references[keyword]}" in converted_words[-1]):
+                cross_reference = f"<a href=\"{hard_coded_references[keyword]}\">{keyword}</a>"
+
+        else:
+            # If it is already processed as cross reference, skip over it.
+            if "<xref" not in words[index-1] and \
+                not (converted_words and f"<xref uid=\"{keyword}" in converted_words[-1]):
+                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
+
+    return cross_reference
```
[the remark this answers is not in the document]
dandhlee: Applied suggestion in separate commit.

docfx_yaml/extension.py
```diff
@@ -1417,29 +1417,94 @@ def find_markdown_pages(app, outdir):
...
+        Empty string if no reference needs to be converted, or a string that contains
+          the converted reference.
+
+    """
+    cross_reference = ""
+    if keyword != current_name and keyword not in current_name and keyword in word:
+        if keyword in hard_coded_references:
+            if "<a" not in words[index-1] and \
+                not (converted_words and f"<a href=\"{hard_coded_references[keyword]}" in converted_words[-1]):
+                cross_reference = f"<a href=\"{hard_coded_references[keyword]}\">{keyword}</a>"
```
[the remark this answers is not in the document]
dandhlee: I thought it was better style to put the return at the end, but I have no preference. Updated to reflect suggestion.

docfx_yaml/extension.py
dandhlee: Done.

dandhlee:
Thank you! Made necessary modifications, updated existing unit test for checking against the name itself and name change in the test. Please take a look again :)

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
+        else:
+            # If it is already processed as cross reference, skip over it.
+            if "<xref" not in words[index-1] and \
+                not (visited_words and f"<xref uid=\"{keyword}" in visited_words[-1]):
+                return f"<xref uid=\"{keyword}\">{keyword}</xref>"
+
+    return None
+
+
+def convert_cross_references(content: str, current_object_name: str, entry_names: List[str]):
```
tbpg, line 1457:
```suggestion
def convert_cross_references(content: str, current_object_name: str, entry_names: List[str]) -> str:
```

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
+        cross_reference = None
         for keyword in entry_names:
-            if keyword != current_name and keyword not in current_name and keyword in word:
-                # If it is already processed as cross reference, skip over it.
-                if "<xref" in words[index-1] or (new_words and f"<xref uid=\"{keyword}" in new_words[-1]):
-                    continue
-                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
-                new_words.append(word.replace(keyword, cross_reference))
+            # Do not convert references to itself.
+            if keyword in current_object_name:
```
tbpg, line 1496: Apologies if this was asked before. Why is this `in` and not `==`? I'm imagining something like `foo.Foo` and `foo.FooBar`. I suppose there might be other punctuation or something?

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
+    Returns:
+        content that has been modified with proper cross references if found.
+
+    """
     words = content.split(" ")
-    new_words = []
+
+    # Contains a list of words that is not a valid reference or converted
+    # references.
+    visited_words = []
```
tbpg, line 1478: What about `processed`, `resolved`, or `converted` `content`? Or even `result`?

(Visited makes me think of doing a graph traversal.)

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
-                # If it is already processed as cross reference, skip over it.
-                if "<xref" in words[index-1] or (new_words and f"<xref uid=\"{keyword}" in new_words[-1]):
-                    continue
-                cross_reference = f"<xref uid=\"{keyword}\">{keyword}</xref>"
-                new_words.append(word.replace(keyword, cross_reference))
+            # Do not convert references to itself.
+            if keyword in current_object_name:
+                continue
+
+            cross_reference = resolve_cross_reference(
```
tbpg, line 1499: :100: This looks a lot better to me.

docfx_yaml/extension.py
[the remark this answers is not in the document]
tbpg: :+1: Please update this docstring. Something like `Returns the resolved cross reference for current_word, if it exists`.

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
+) -> Optional[str]:
+    """Given a word, check if it contains a reference keyword we should convert.
 
-# Finds and replaces occurrences which should be a cross reference in the given
-# content, except for the current name.
-def convert_cross_references(content: str, current_name: str, entry_names: List[str]):
+    Args:
+        words: list of words used to check and compare content before and after `current_word`
+        index: index position of `current_word` within words
+        current_word: current word being looked at
```
tbpg, line 1433: Total nit: perhaps `current_word` should be the first arg?

docfx_yaml/extension.py
```diff
@@ -1417,58 +1417,127 @@ def find_markdown_pages(app, outdir):
...
+        visited_words: list of words containing words that's been processed so far
+        hard_coded_references: Optional list containing a list of hard coded reference
+
+    Returns:
+        None if current word does not contain a reference keyword, or a string
+          that contains the converted reference.
+    """
+    if keyword in current_word:
+        if hard_coded_references and keyword in hard_coded_references:
+            if "<a" not in words[index-1] and \
```
tbpg, line 1444: Will you add a comment to this? I don't understand why `<a` being in the previous word makes a difference.

Same with the check on the next line.

tbpg (changes requested):
Sorry for all the comments -- trying to wrap my head around this. :smile:

docfx_yaml/extension.py

dandhlee (commented):
No worries, thank you for taking a detailed look! :D

tbpg (dismissed):

dandhlee (commented):
Made some additional modifications, please take a look again!

tbpg (approved):

PR #222, opened 2022-07-11
feat: support devsite notices

dandhlee:
Adds support for notices such as `note`, `warning` and more, to be rendered for usage in cloud.google.com.

Reuses a lot of the code from other docstring parser. There must be a way I can refactor this for generic parser but I couldn't think of it at the moment, I'll work on that in the future.

Updated a unit test for a type that's not supported yet (`literalincludes`), and added tests for several types of notices.

Staged and verified that it works as intended.

Towards b/206698969.

- [x] Tests pass


tbpg (approved):
LGTM. We may want special handling/styling for deprecated/removed items for all languages in the future.

PR #226, opened 2022-07-18
ci: skip empty packages

### LINKED ISSUE #225: generate-docs Kokoro job not handling empty blobs properly
There seems to be empty blobs that are able to make its way to the GCS bucket. The job should check if there were any files produced in `docs/_build/html/docfx_yaml` and then not process the blob.

As a followup, perhaps it might be a good idea to check why it might not have produced any content. This could be desired (broken version released) or there is an actual error that I should look into.

One simple idea could be to print all the blobs that have not produced any content in the logs.

### LINKED ISSUE #250: generate: docfx-python-redis-1.0.1.tar.gz failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/main/packages/flakybot).

If I'm commenting on this issue too often, add the `flakybot: quiet` label and
I will stop commenting.

---

commit: 48f1842f3b0d9af10dc09c10d8864233056129c1
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/3e2e3f1a-a872-4731-9ea0-788c88cc25f8), [Sponge](http://sponge2/3e2e3f1a-a872-4731-9ea0-788c88cc25f8)
status: failed

dandhlee:
After running sphinx to build documentation, there wasn't any check to ensure that there is content produced. This caused issues like empty tarballs being passed and failing to build in the pipeline.

Adding in a check to see if any documentation has been produced by checking the contents of `docs/_build/html/docfx_yaml`, which is the output directory (In the near future I'll refactor this to make it a variable as `output_directory`). Simple check for `ls | wc -l` will suffice to check whether there's content in the directory or not.

Also adding in `empty_packages` to log the packages that did not produce any documentation; maybe this is expected or unexpected, but having this logged at the very end should help track things after running the script.

Fixes #225.
Fixes https://github.com/googleapis/doc-pipeline/issues/250.

- [x] Tests pass

### DIFF
--- a/.kokoro/generate-docs.sh
+++ b/.kokoro/generate-docs.sh
@@ -119,7 +119,7 @@ for package in $(echo "${python_bucket_items}" | cut -d "-" -f 5- | rev | cut -d
 
     # Check that documentation is produced. If not, log and continue.
     file_num=$(ls docs/_build/html/docfx_yaml/ | wc -l)
-    if [ ${file_num} -eq 0]; then
+    if [ ${file_num} -eq 0 ]; then
       empty_packages="${repo}-${tag} ${empty_packages}"
       continue
     fi
--- a/.kokoro/generate-docs.sh
+++ b/.kokoro/generate-docs.sh
@@ -118,8 +118,7 @@ for package in $(echo "${python_bucket_items}" | cut -d "-" -f 5- | rev | cut -d
     nox -s docfx
 
     # Check that documentation is produced. If not, log and continue.
-    file_num=$(ls docs/_build/html/docfx_yaml/ | wc -l)
-    if [ ${file_num} -eq 0 ]; then
+    if [ ! "$(ls docs/_build/html/docfx_yaml/)" ]; then
       empty_packages="${repo}-${tag} ${empty_packages}"
       continue
     fi
--- a/.kokoro/generate-docs.sh
+++ b/.kokoro/generate-docs.sh
@@ -38,6 +38,8 @@ python3 -m pip install --user django==2.2 ipython
 
 # Store the contents of bucket log in a variable to reuse.
 python_bucket_items=$(gsutil ls "gs://docs-staging-v2/docfx-python*")
+# Store empty tarballs that did not produce any content to check later.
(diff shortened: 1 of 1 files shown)


dandhlee:
The test successfully produced the following output:
```
The following packages did not produce any content:
python-websecurityscanner-v0.1.0
python-websecurityscanner-v0.1.1
python-vision-v0.21.0
python-vision-v0.22.0
python-vision-v0.23.0
```
as expected!

.kokoro/generate-docs.sh
```diff
@@ -115,6 +117,13 @@ for package in $(echo "${python_bucket_items}" | cut -d "-" -f 5- | rev | cut -d
     # Build YAML tarballs for Cloud-RAD.
     nox -s docfx
 
+    # Check that documentation is produced. If not, log and continue.
+    file_num=$(ls docs/_build/html/docfx_yaml/ | wc -l)
+    if [ ${file_num} -eq 0 ]; then
```
tbpg, line 122:
```suggestion
    if [ ! "$(ls docs/_build/html/docfx_yaml/)" ]; then
```

.kokoro/generate-docs.sh
```diff
@@ -38,6 +38,8 @@ python3 -m pip install --user django==2.2 ipython
 
 # Store the contents of bucket log in a variable to reuse.
 python_bucket_items=$(gsutil ls "gs://docs-staging-v2/docfx-python*")
+# Store empty tarballs that did not produce any content to check later.
```
tbpg, line 41: Design question: why not do this in the plugin? This seems like an error we should be notified about?

tbpg (changes requested):

.kokoro/generate-docs.sh
dandhlee: Filed https://github.com/googleapis/docuploader/issues/114 to handle the issue on docuploader as well.

dandhlee (commented):
Please take a look again!

tbpg (approved):

PR #255, opened 2022-10-05
feat: use GitHub's default README as index page

### LINKED ISSUE #105: Parse index.md files into index.yaml
Googlers see b/200623336.

The autogenerated `index.md` files are quite useful entry pages into each client libraries. However since `index.yaml` already exists from the plugin, we should merge the content of the two together.

### LINKED ISSUE #107: Rename README files into another name.
Googlers see b/206694500.

Cloud site does not allow files to be named `README`, will need to rename this file to something else to make use of it.

dandhlee:
Instead of using a custom index page that provided little to no benefit, this PR is incorporating the existing README as the main index page. This was asked by a lot of DPEs for their client libraries' landing page.

The markdown generator unfortunately messed around with the spacing for image links (see tests/markdown_example_bad_image_links.md file for reference), but all other parts I've verified is looking good. (See staged version of the doc for https://cloud.google.com/python/docs/reference/storage/latest). Adding a small parser to help fix that for the index file.

Fixes #105.
Fixes #107.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1429,6 +1429,35 @@ def highlight_md_codeblocks(mdfile):
         mdfile_iterator.write(new_content)
 
 
+def clean_image_links(mdfile_path: str) -> None:
+    """Cleans extra whitespace that breaks image links in index.html file."""
+    image_link_pattern='\[\s*!\[image\]\(.*\)\s*\]\(.*\)'
+    new_lines = []
+    with open(mdfile_path) as mdfile_iterator:
+        file_content = mdfile_iterator.read()
+        broken_image_links = [
+            [m.start(), m.end()]
+            for m in re.finditer(image_link_pattern, file_content)
+        ]
+
+        prev_start = prev_end = 0
+
+        for start, end in broken_image_links:
+            matched_str = file_content[start:end]
+            # Clean up all whitespaces for the image link.
+            clean_str = ''.join(matched_str.split())
+
+            new_lines.append(file_content[prev_end:start])
+            new_lines.append(clean_str)
+            prev_start, prev_end = start, end
+
+        new_lines.append(file_content[prev_end:])
+
+    with open(mdfile_path, 'w') as mdfile_iterator:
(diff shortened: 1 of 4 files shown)


docfx_yaml/extension.py
```diff
@@ -1482,12 +1511,22 @@ def find_markdown_pages(app, outdir):
...
                     prepend_markdown_header(name, mdfile_iterator)
 
-            shutil.copy(mdfile, f"{outdir}/{mdfile.name.lower()}")
+            mdfile_name_to_use = mdfile.name.lower()
+            if mdfile_name_to_use in files_to_rename:
+                mdfile_name_to_use = files_to_rename[mdfile_name_to_use]
+
+            shutil.copy(mdfile, f"{outdir}/{mdfile_name_to_use}")
+
+            # Do not add index file to the TOC.
```
tbpg, line 1520: Why not? Seems like we could set `name` to `Overview` here and not need the special handling down below (`[{'name': 'Overview', 'href': 'index.md'}] + app.env.markdown_pages + pkg_toc_yaml`)?

Perhaps for sorting? Still seems like it would be better to handle all md pages in this function.

docfx_yaml/extension.py
```diff
@@ -1482,12 +1511,22 @@ def find_markdown_pages(app, outdir):
...
+            if mdfile_name_to_use == 'index.md':
+                # Clean up any broken image links for the index file.
+                clean_image_links(f"{outdir}/{mdfile_name_to_use}")
```
tbpg, line 1523: This doesn't apply for every md file?

tests/markdown_example_bad_image_links.md
```diff
@@ -0,0 +1,23 @@
+# Python Client for Google Cloud Storage API
+
+[
+
+![image](https://img.shields.io/badge/support-stable-gold.svg)
```
tbpg, line 5: Seems like GitHub doesn't render these properly?

dandhlee (commented):

tbpg (approved):

PR #258, opened 2022-10-07
fix: render emphasis properly and markdown in tables

dandhlee:
Cloud site does not recognize the double backtick (` `` `) being used as providing emphasis. To get this to be recognized, I've parsed content to render back to single backtick which is indeed recognized by the Cloud site tenant. There might be more items I may need to format down the road, so I've made a generic method and will expand as needed.

However, Markdown syntax within HTML tables (within Responsive tables) such as codeblocks (`` ``` ``) and the single backtick for emphasis does not get rendered properly. To accommodate this, I've created a method that takes care of all Markdown -> HTML related reformatting, starting with just the two that I've noticed.

Towards b/248339091 and b/238462759.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -83,12 +83,12 @@ REFMETHOD = 'meth'
 REFFUNCTION = 'func'
 INITPY = '__init__.py'
 # Regex expression for checking references of pattern like ":class:`~package_v1.module`"
-REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.\<\> ]*(\(\))?`'
+REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_.<> ]*(\(\))?`'
 # Regex expression for checking references of pattern like "~package_v1.subpackage.module"
-REF_PATTERN_LAST = '~([a-zA-Z0-9_\<\>]*\.)*[a-zA-Z0-9_\<\>]*(\(\))?'
+REF_PATTERN_LAST = '~([a-zA-Z0-9_<>]*\.)*[a-zA-Z0-9_<>]*(\(\))?'
 # Regex expression for checking references of pattern like
 # "[module][google.cloud.cloudkms_v1.module]"
-REF_PATTERN_BRACKETS = '\[[a-zA-Z0-9\_\<\>\-\. ]+\]\[[a-zA-Z0-9\_\<\>\-\. ]+\]'
+REF_PATTERN_BRACKETS = '\[[a-zA-Z0-9_<>\-. ]+\]\[[a-zA-Z0-9_<>\-. ]+\]'
 
 REF_PATTERNS = [
     REF_PATTERN,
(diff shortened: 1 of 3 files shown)


docfx_yaml/extension.py
```diff
@@ -758,12 +758,107 @@ def _extract_docstring_info(summary_info, summary, name):
...
+            reformatted_lines.append(line)
+
+    return '\n'.join(reformatted_lines)
+
+
+def _reformat_codeblocks(content: str) -> str:
+    """Formats codeblocks from ``` to <pre>."""
+    codeblock_fence = '```'
+    codeblock_current = '<pre>'
+    codeblock_next = '</pre>'
```
tbpg, line 805: Might just be me, but I find the `codeblock` prefix kind of confusing in this. We're in the `codeblock` function, why do we need this prefix? And I'd expect a different prefix for the fence than the HTML tags.

Especially when the there is a `triple_backtick` variable above. That name is clear to me.

docfx_yaml/extension.py
```diff
@@ -758,12 +758,107 @@ def _extract_docstring_info(summary_info, summary, name):
...
         else:
             words.append(word)
-    
+
         index += 1
 
     return top_summary
 
 
+def reformat_summary(summary: str) -> str:
```
tbpg, line 767: Consider moving this down below the other "exported" reformat function.

docfx_yaml/extension.py
```diff
@@ -83,9 +83,9 @@ class Bcolors:
 REFFUNCTION = 'func'
 INITPY = '__init__.py'
 # Regex expression for checking references of pattern like ":class:`~package_v1.module`"
-REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.<> ]*(\(\))?`'
+REF_PATTERN = ':(py:)?(func|class|meth|mod|ref|attr|exc):`~?[a-zA-Z0-9_\.\<\> ]*(\(\))?`'
```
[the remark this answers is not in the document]
dandhlee: This format appears in a lot of documentation, mostly from hand-written comments as well. It's easier to accommodate these and other sphinx-specific format as well, so folks can freely explore the Sphinx functionalities and I could follow up to add support for them if they turn out to be missing.

This would take a lot of time to clean up on hand-written libraries that are less maintained as well, so I think it's more efficient to handle it in the plugin.

docfx_yaml/extension.py

docfx_yaml/extension.py
```diff
@@ -758,12 +758,107 @@ def _extract_docstring_info(summary_info, summary, name):
...
 
     return top_summary
 
 
+def reformat_summary(summary: str) -> str:
+    """Applies any style changes to be made specifically for DocFX YAML.
+
+    Makes the following changes:
+      - converts ``text`` to `text`
+      - [add any additional changes to be made]
```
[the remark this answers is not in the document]

docfx_yaml/extension.py
```diff
@@ -758,12 +758,107 @@ def _extract_docstring_info(summary_info, summary, name):
...
+    double_backtick = '``'
+    triple_backtick = '```'
+    for line in summary.split('\n'):
+        # Check that we're only looking for double backtick (``) and not
+        # comments (```).
+        if triple_backtick not in line and double_backtick in line:
+            reformatted_lines.append(line.replace(double_backtick, single_backtick))
+
+        # Add any additional changes to be made here.
+        #elif:
```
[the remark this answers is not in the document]

docfx_yaml/extension.py
dandhlee: Agreed - I shouldn't have to go back and forth between the names. Done.

dandhlee (commented):

tbpg (approved):

Issue #115, opened 2021-08-25
Refactor tests into utility unit tests and bigger unit tests


dandhlee:
Googlers see b/206698957.

For functions that are small utility functions, they should be in a separate unit test file since `tests/test_unit.py` is getting rather large. Will keep using `tests/test_unit.py` for bigger tests for stuff like checking docstring parsing, but not for utility functions like `find_package_group`.

dandhlee:
In `tox.ini` file, instead of using `python3 -m unittest tests/test_unit.py tests/test_helpers.py tests/test_markdown.py` to list out all the files, try https://docs.python.org/3/library/unittest.html#test-discovery.

PR #259, opened 2022-10-10
refactor: move markdown into its own file & testing

dandhlee:
Taking Markdown functionality out of the extension file. The file is way too big and should be refactored further.

Updated docstring for all of the Markdown methods as well.

Moved the testing into its own file, however I've temporarily included unit tests for private methods as well, as we don't have goldens testing yet at the moment. Goldens testing will follow next, before more refactoring can be done to ensure the refactoring doesn't actually change the output format.

Updates #115. Towards #44.

- [x] Tests pass


docfx_yaml/extension.py
```diff
@@ -51,6 +51,9 @@
 from .monkeypatch import patch_docfields
 from .directives import RemarksDirective, TodoDirective
 from .nodes import remarks
+from .markdown_utils import run_sphinx_markdown
+from .markdown_utils import reformat_markdown_to_html
+from .markdown_utils import move_markdown_pages
```
tbpg, line 56: I think style says to import the module, rather than the individual methods/classes within it.

tox.ini
```diff
@@ -22,7 +22,7 @@ commands =
 deps =
     {[testenv]deps}
 commands =
-    python3 -m unittest tests/test_unit.py tests/test_helpers.py
+    python3 -m unittest tests/test_unit.py tests/test_helpers.py tests/test_markdown.py
```
tbpg, line 25: Is there a way to run all tests in the `tests` directory rather than having to list out each file? See https://docs.python.org/3/library/unittest.html#test-discovery. Perhaps:

```
python3 -m unittest discover tests
```

dandhlee: I've left a note on https://github.com/googleapis/sphinx-docfx-yaml/issues/115 to look into this in a followup PR. There's a `test_yaml.py` in the test directory that's not making this easy to discover just set select examples.

docfx_yaml/extension.py
dandhlee: Oops, updated. Done.

docfx_yaml/extension.py
```diff
@@ -51,6 +51,7 @@
 from .monkeypatch import patch_docfields
 from .directives import RemarksDirective, TodoDirective
 from .nodes import remarks
```
dandhlee, line 53: These imports will be refactored in a followup, and clean up methods that are not used.

dandhlee:

tbpg (approved):

PR #261, opened 2022-10-31
chore(deps): update dependencies

### LINKED ISSUE #260: Add "google-cloud-aiplatform[prediction]" as custom requirement for Kokoro script
There is a new library in python-aiplatform that requires a special install. Filing the issue to keep track of this.

dandhlee:
Adding specific installation for `aiplatform[prediction]`. Also making the `requirements.in` file look nicer :)

Fixes #260.

- [x] Tests pass


dandhlee:
Confirmed `prediction` is picked up from the update based on this PR:
```
writing output... [ 98%] prediction_v1/types
writing output... [100%] prediction_v1beta1/types
```
this would not show if `prediction` library is not installed as Sphinx would not be able to pick up on it.

tbpg (commented):
Why can't the library itself install these dependencies?

dandhlee:
I have learned what they are! They are called library extras, when they're installed as like `aiplatform[prediction]` with extras being inside `[]`. They are optional extras that folks can install.

Dustin filed a feature request before: https://github.com/pypa/setuptools/pull/1503, but tl;dr is that with current state of python packaging, it is not possible to ask for users to also install all or some extras as default when you simply install `aiplatform` for example.

dandhlee:
Ah, they can! I'm just adding support for this in the mega generator in case I have to regenerate all the docs and didn't want to miss out on `prediction` or other smaller bits as well.

tbpg:
Will you add a comment explaining that? Perhaps that regeneration script could install those dependencies separately?

dandhlee:
Done! Please take a look again.

tbpg (approved):

PR #265, opened 2022-11-01
fix: remove unused YAML fields to fix presubmit

### LINKED ISSUE #264: Presubmit failing for specifying branch on main
Branch parameter really isn't used anywhere. This could be taken out.

dandhlee:
The `remote` settings are not consumed for c.g.c. usage or anywhere else. Removing it in the meanwhile because this discrepancy is causing test failures and also causing local test failure for repo.

Fixes #264.

- [x] Tests pass


tbpg (approved):
Unfortunate to remove this since I think we'll want it eventually for source links. But, makes sense for now.

PR #266, opened 2022-11-14
feat!: retrieve all markdown pages

### LINKED ISSUE #217: Missing markdown pages in directories
The plugin assumed that all markdown pages exist under one flat directory within `docs/` folder usually. However this is not the case and we're now missing documentation as a result of this assumption.

Example: https://cloud.google.com/python/docs/reference/bigquery/latest/ is missing all pages under https://github.com/googleapis/python-bigquery/tree/main/docs/usage.

dandhlee:
Recovered all markdown pages in nested directory structures. This was not correctly retrieved as I was only looking in the top directory.

Few things that's happening (happy to break down, but was hard to do so to have all the moving parts in one picture):
* Keeping track of where pages are retrieved. This is vital as we need to know where markdown pages will need to end up in the TOC
* Merging the Markdown and Package's TOC, based on the two sources pulled separately. I could choose to just move all the markdown files at the root directory, but I think a better experience will be to put the markdown pages where they are located.
* Remove unused markdown pages that are generated. These are redundant API pages that Sphinx generated but are not used by DocFX Yaml plugin.

Only the `handwritten` TOC is updated and works as expected, `gapic-combo` and `gapic-auto` doesn't contain any markdown pages in nested directories.

Some markdown pages might come with a mix of handwritten guides and autogenerated API content. I plan on working with the library maintainers to have this updated throughout (<10) for the purpose of this plugin that separates Sphinx HTML and DocFX YAML content.

**Breaking Change**: Upgrading to using 3.9. All Python client libraries are able to use 3.9, so this is not a problem, but will be a backwards incompatible change.

Fixes #217.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -26,6 +26,8 @@ import shutil
 import black
 import logging
 
+from collections import defaultdict
+from collections.abc import MutableSet
 from pathlib import Path
 from functools import partial
 from itertools import zip_longest
@@ -160,8 +162,11 @@ def build_init(app):
     app.env.docfx_uid_names = {}
     # This stores file path for class when inspect cannot retrieve file path
     app.env.docfx_class_paths = {}
-    # This stores the name and href of the markdown pages.
-    app.env.markdown_pages = []
+    # This stores the name and href of the nested markdown pages.
+    app.env.markdown_pages = defaultdict(list)
+    # This stores all the markdown pages moved from the plugin, will be used
+    # to compare and delete unused pages.
+    app.env.moved_markdown_pages = set()
 
     app.env.docfx_xrefs = {}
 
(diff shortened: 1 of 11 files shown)


docfx_yaml/markdown_utils.py
```diff
@@ -254,7 +259,14 @@ def move_markdown_pages(app: sphinx.application, outdir: Path) -> None:
         'readme.md': 'index.md',
     }
 
-    markdown_dir = Path(app.builder.outdir).parent / "markdown"
+    base_markdown_dir = Path(app.builder.outdir).parent / "markdown"
+
+    markdown_dir = (
+        base_markdown_dir / '/'.join(cwd)
```
tbpg, line 265:
```suggestion
        base_markdown_dir.joinpath(*cwd)
```

(Using `'/'` and `pathlib` at the same time is a red flag.)

tests/testdata/goldens/handwritten/toc.yml
```diff
@@ -4,6 +4,16 @@
   - href: changelog.md
     name: Changelog
   - items:
+    - href: blobs.md
```
tbpg, line 7: Do these test files need to be as long as they are?

tbpg (changes requested):

docfx_yaml/extension.py
```diff
@@ -1547,6 +1552,68 @@ def search_cross_references(obj, current_object_name: str, known_uids: List[str]
...
+    Returns:
+        A set of markdown pages that has been added, and the merged table of
+        contents file, with files in the correct position.
+    """
+    added_pages = set()
+
+    pkg_toc_queue = [package for package in pkg_toc_yaml]
+    for entry in pkg_toc_queue:
+        if (children := entry.get('items')):
+            pkg_toc_queue.extend(children)
```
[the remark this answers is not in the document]
dandhlee: Done. Added `_flatten_tocs` entry to address this.

dandhlee (commented):

docfx_yaml/extension.py
```diff
@@ -1547,6 +1553,78 @@ def search_cross_references(obj, current_object_name: str, known_uids: List[str]
...
+    """
+    def _flatten_toc(
+        toc_yaml_entry: list[_toc_yaml_type_alias],
+    ) -> list[_toc_yaml_type_alias]:
+        """Flattens and retrieves all children within pkg_toc_yaml."""
+        toc_queue = list(toc_yaml_entry)
+        for entry in toc_queue:
+            if (children := entry.get('items')):
+                toc_queue.extend(_flatten_toc(children))
+                toc_queue.extend(children)
```
tbpg, line 1585: This is still modifying `toc_queue` while looping over it. Can you iterate over `toc_yaml_entry` instead?

If I'm understanding correctly, we don't need a queue at all. Just need to iterate over the "graph" and put it all into a single list.

dandhlee: Oops 🤦 Made updates.

dandhlee:
Please take a look again!

docfx_yaml/extension.py
```diff
@@ -1527,6 +1533,78 @@ def search_cross_references(obj, current_object_name: str, known_uids: List[str]
...
+    """
+    def _flatten_toc(
+        toc_yaml_entry: list[_toc_yaml_type_alias],
+    ) -> list[_toc_yaml_type_alias]:
+        """Flattens and retrieves all children within pkg_toc_yaml."""
+        entries = list(toc_yaml_entry)
+        for entry in toc_yaml_entry:
+            if (children := entry.get('items')):
+                entries.extend(_flatten_toc(children))
+                entries.extend(children)
```
tbpg, line 1565: Does this line end up double-adding? `_flatten_toc` is called with `children`, which will then return `children` (and more), then we add `children` again right here.

Seems like this needs a test.

dandhlee: 🤦🤦 Yes you're right. Removed the redundant line, added unit test coverage as well. I didn't add it before because it required `sphinx.app` ... before, but not anymore so fine to add now.

dandhlee:
Please take a look again!

tbpg (approved):

PR #280, opened 2023-02-03
fix: allow multiple/custom index pages to be used

### LINKED ISSUE #269: TOC not being generated properly for pandas-gbq
I've noticed that with the recent updates the TOC looks deformed.

Previously: 

```
- items:
  - href: privacy.md
    name: Privacy
  - href: contributing.md
    name: Contributing to pandas-gbq
  - href: changelog.md
    name: Changelog
  - href: writing.md
    name: Writing Tables
  - href: intro.md
    name: Introduction
  - href: api.md
    name: API Reference
  - href: install.md
    name: Installation
  - href: reading.md
    name: Reading Tables
  - items:
    - name: Context
      uid: pandas_gbq.Context
    name: Context
  name: pandas-gbq
```

with newest sphinx plugin version:
```
- items:
  - - items:
      - name: Context
        uid: pandas_gbq.Context
      name: Context
  name: pandas-gbq
```
Seems like missing the index pages is indeed a problem. :/

dandhlee:
Investigate this on a non-cloud library not generating the TOC properly due to missing a README page. However, upon further research, some libraries also used multiple `index.rst` pages nested. Solving both issues by allowing (1) `index.rst` to be used in place of `README.rst` if `README.rst` is not present so TOC is well formed and still has an overview page, and (2) allow multiple index pages to be present, just under a different name.

Added a `non-cloud` goldens section to help cover for the additional coverage as no unit test will be available for this specific portion of the code, covered by the goldens. Non Cloud libraries should still be added to the coverage, some time or later when we support non-Cloud libraries as well, so I'm adding them now.

No changes to existing Cloud client libraries and their TOC generation.

Fixes #269.

- [x] Tests pass


docfx_yaml/markdown_utils.py
```diff
@@ -301,6 +301,16 @@ def move_markdown_pages(
     # Used to keep track of the index page entry to insert later.
     index_page_entry = None
 
+    list_of_markdown_file_names = [
```
tbpg, line 304: The type of this variable is a list. No need to have list in the name. So, you can just call this `markdown_file_names`.

docfx_yaml/markdown_utils.py
```diff
@@ -301,6 +301,16 @@ def move_markdown_pages(
...
+        mdfile.name.lower()
+        for mdfile in markdown_dir.iterdir()
+    ]
+
+    # If a custom index.md file is preferred, use this instead of README.md.
```
tbpg, line 309:
```suggestion
    # If there is an index.md and no readme.md, use the index.md. Otherwise, we ignore the index.md.
```

docfx_yaml/markdown_utils.py
tbpg, line 304: Use a set instead of a list since this is used to test existence and order doesn't matter?

dandhlee:
Oops, region tags shouldn't be added! Stripped them off the files (they're needed as some code snippets were pulled in from Sphinx configs)

docfx_yaml/markdown_utils.py
dandhlee: Done.

dandhlee: Done.

dandhlee (commented):
Updated the region tags slightly so Sphinx continue to work with `literalinclude` but also doesn't trigger snippetbot.

Thank you for the review! Please take a look again :D

tbpg (approved):

PR #289, opened 2023-02-16
build: update generate-docs to support monorepo

### LINKED ISSUE #288: Update generate-docs to handle google-cloud-python monorepo
`generate-docs` script was built with split repos in mind, but does not work with monorepo containing packages under `google-cloud-python/packages`. This hinders the ability to regenerate documentation for those monorepo libraries.

dandhlee:
Adds support for generating documentation for [`google-cloud-python`](https://github.com/googleapis/google-cloud-python) repository. For corresponding packages already in the pipeline, crawls and regenerates documentation as required for those in the monorepo.

In case additional support is required for monorepo, instead of tweaking the existing code slightly I've added another section to handle monorepo portion separately. This is also likely required for extracting previous release tags within the monorepo, which will be much more complicated to handle. For now, I'll focus on supporting latest releases for the monorepo.

I've verified locally that the script works as expected - reproduced 6 packages found in the current pipeline for 6 Cloud packages found in https://github.com/googleapis/google-cloud-python/tree/main/packages. Checked that the script works with Kokoro test run.

Fixes #288.
Towards b/249115548.

- [x] Tests pass


.kokoro/generate-docs.sh
```diff
@@ -55,6 +57,14 @@ for package in $(echo "${python_bucket_items}" | cut -d "-" -f 5- | rev | cut -d
   tar -zxvf ${tarball}
   repo=$(cat docs.metadata | grep "github_repository:" | cut -d "\"" -f 2 | cut -d "/" -f 2)
 
+  # If the pacakage is part of the monorepo, we'll process this later.
+  if [[ "${repo}" == "google-cloud-python" ]]; then
+    monorepo_packages="$(cat docs.metadata | grep "distribution_name" | cut -d "\"" -f 2) ${monorepo_packages}"
```
tbpg, line 62: Why the `${monorepo_packages}` at the end?

.kokoro/generate-docs.sh
```diff
@@ -152,7 +162,57 @@ for package in $(echo "${python_bucket_items}" | cut -d "-" -f 5- | rev | cut -d
...
 
-if [ ! ${empty_packages} ]; then
+# Build documentation for monorepo packages
+if [ -n "${monorepo_packages}" ]; then
+  echo "Processing monorepo packages"
+  git clone "https://github.com/googleapis/google-cloud-python.git"
+  cd google-cloud-python/packages
+
+  # TODO (https://github.com/googleapis/sphinx-docfx-yaml/issues/287): support
+  # mutl-version build for the monorepo.
```
tbpg, line 172: multi

tbpg (approved):

.kokoro/generate-docs.sh
dandhlee: 🤦  Thank you!

.kokoro/generate-docs.sh
dandhlee: To continue adding to the string to be processed later as I'm keeping it in a string format, which then gets used at line 172+ to process monorepo-specific packages.

dandhlee: Didn't know you could also use `+=` in bash as well. Updated to append with a whitespace at the end.

PR #311, opened 2023-08-22
fix: support docstrings without type added

dandhlee:
Previously, GoogleDocstrings from Napoleon required that docstrings came with a type hint. However, docstrings in the updated Python style guide do not come with type hint and that should be supported.

Using the updated extract_type_name method, uses the default type hint if they're not found in the docstring. Also fixing `object at 0x` to be simpler to catch more cases.

Fixes b/296938464.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -740,22 +740,20 @@ def _extract_type_name(annotation: Any) -> str:
     annotation.
 
     Args:
-        annotation: the inspected object in its type format.
+        annotation: the inspected object in its type format. The type hint used
+            is `Any`, because it's the type of the object inspected itself,
+            which can come as any type available.
 
     Returns:
         The extracted type hint in human-readable string format.
     """
-    type_name = ""
-    # Extract names for simple types.
-    try:
+
+    annotation_dir = dir(annotation)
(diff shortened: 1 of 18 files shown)


docfx_yaml/extension.py
```diff
@@ -845,9 +845,13 @@ def _update_friendly_package_name(path):
...
 
-                        # Only add defaultValue when str(default) doesn't contain object address string(object at 0x)
-                        # inspect.getargspec method will return wrong defaults which contain object address for some default values, like sys.stdout
-                        if 'object at 0x' not in str(default):
+                        # Only add defaultValue when str(default) doesn't
+                        # contain object address string, for example:
+                        # (object at 0x) or <lambda> at 0x7fed4d57b5e0
+                        # inspect.getargspec method will return wrong defaults
+                        # which contain object address for some default values,
+                        # like sys.stdout.
```
tbpg, line 853: I feel like there is a missing period or "The" or something here. I don't understand this comment.

tbpg (approved):

docfx_yaml/extension.py
dandhlee: Updated the docstring!

PR #312, opened 2023-08-22
fix: refactor type hint retrieval

dandhlee:
Refactors type hint retrieval when inspecting the source code. The indentation was very ugly and finally got to get around fixing it.

Given the annotation, tries to retrieve its name. For more complex types, removes ForwardRef references as they're used for type hinting, not for the user to read.

Updated goldens to verify the fix can be found in #311 - I split this PR into two for smaller PRs to review. Added goldens in this PR seem to not have been picked up from a previous PR, perhaps.

Towards b/296938464. Unblocks #311.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -740,7 +740,9 @@ def _extract_type_name(annotation: Any) -> str:
     annotation.
 
     Args:
-        annotation: the inspected object in its type format.
+        annotation: the inspected object in its type format. The type hint used
+            is `Any`, because it's the type of the object inspected itself,
+            which can come as any type available.
 
     Returns:
         The extracted type hint in human-readable string format.
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -734,11 +734,14 @@ def extract_product_name(name):
 
 def _extract_type_name(annotation: Any) -> str:
     """Extracts the type name for the given inspected object.
+
     Used to identify and extract the type hints given through inspecting the
     source code. Carefully extracts only the relevant part for the given
     annotation.
+
     Args:
         annotation: the inspected object in its type format.
+
     Returns:
         The extracted type hint in human-readable string format.
     """
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -758,11 +758,11 @@ def _extract_type_name(annotation: Any) -> str:
         return type_name
 
     # If ForwardRef references are found, recursively remove them.
-    prefix_to_remove_start = "[ForwardRef('"
+    prefix_to_remove_start = "ForwardRef('"
     if prefix_to_remove_start not in type_name:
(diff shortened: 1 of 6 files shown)


docfx_yaml/extension.py
```diff
@@ -732,6 +732,52 @@ def extract_product_name(name):
     return product_name
 
 
+def _extract_type_name(annotation: Any) -> str:
```
dansaadati, line 735: Are annotations and type hints the same? If not, I'd rename `annotation` to `type_hint`, or maybe `argument_annotations`.  Feel free to Ack if this breaks your current naming conventions.

docfx_yaml/extension.py
```diff
@@ -732,6 +732,52 @@ def extract_product_name(name):
...
+    while prefix_to_remove_start in type_name:
+        start_index = type_name.find(prefix_to_remove_start)
+        end_index = type_name.find(prefix_to_remove_end, start_index)
+        type_name = ''.join([
+            type_name[:start_index],
+            type_name[start_index+prefix_start_len:end_index],
+            type_name[end_index+prefix_end_len:],
+        ])
+
+    return type_name
```
dansaadati, line 778: Please consider adding a unit test for this new logic.

docfx_yaml/extension.py
dansaadati, line 735: `Dict[Any]`?

dansaadati (approved):

docfx_yaml/extension.py
dandhlee: I don't think it's quite the same - annotations may hint at the type hint, however it's more like the object that includes the type hint. Probably best to stick to `annotations` from https://docs.python.org/3/library/inspect.html#inspect.get_annotations.

dandhlee: Ah, tried looking into the type and it's very messy lol it returns the inspected object's class directly. I will add this to the docstring!

FYI this is what I got from a small sample:
```
type: <class 'typing._CallableGenericAlias'>
type: <class 'typing._GenericAlias'>
type: <class 'typing._CallableGenericAlias'>
type: <class 'typing._GenericAlias'>
type: <class 'typing._CallableGenericAlias'>
type: <class 'typing._GenericAlias'>
type: <class 'typing._CallableGenericAlias'>
type: <class 'typing._GenericAlias'>
type: <class 'typing._CallableGenericAlias'>
type: <class 'typing._UnionGenericAlias'>
type: <class 'typing._CallableType'>
type: <class 'str'>
```

docfx_yaml/extension.py
dandhlee: Let me add a unit test as a followup PR. Filed https://github.com/googleapis/sphinx-docfx-yaml/issues/318.

dandhlee (commented):
Thank you for the review!

Issue #331, opened 2023-10-31
Convert cross references properly for code content


dandhlee:
Currently cross references are only converted to xref tags and are not usable within code content. For example:

```
>>> import <xref uid="bigframes.pandas">bigframes.pandas</xref> as bpd
>>> bpd.options.display.progress_bar = None
```

We'll need to find direct links, however should be possible. See go examples: https://cloud.google.com/go/docs/reference/cloud.google.com/go/alloydb/latest/apiv1#cloud_google_com_go_alloydb_apiv1_DefaultAuthScopes

PR #332, opened 2023-10-31
fix: ignore cross reference conversion in examples

dandhlee:
Bigframes team saw that cross references were converting their example code content, see https://cloud.google.com/python/docs/reference/bigframes/latest/bigframes.session.Session#bigframes_session_Session_read_gbq. The conversion for cross references don't work in code content, temporarily fixing this for Bigframes team's example. 

I've filed #331 to follow up to ensure that code content gets properly converted into links. Until then, for example content found in the summary, they won't get converted into xrefs.

Fixes b/306420925.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -1395,7 +1395,14 @@ def find_uid_to_convert(
     return None
 
 
-def convert_cross_references(content: str, current_object_name: str, known_uids: List[str]) -> str:
+# TODO(https://github.com/googleapis/sphinx-docfx-yaml/issues/331): Improve
+# converting cross references for code content.
+def convert_cross_references(
+    content: str,
+    current_object_name: str,
+    known_uids: List[str],
+    ignore_examples: Optional[bool] = False,
+) -> str:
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -1428,19 +1438,26 @@ def convert_cross_references(content: str, current_object_name: str, known_uids:
...
+            example_index = index
         uid = find_uid_to_convert(
             word, words, index, known_uids, current_object_name, processed_words, hard_coded_references
         )
 
-        if uid:
+        # If the reference is found after example section, ignore it.
+        if uid and (
+            not ignore_examples or
+            (ignore_examples and index < example_index)
```
dansaadati, line 1453: LGTM, as long as there cannot be content after the example.

dansaadati (approved):
-

docfx_yaml/extension.py
dandhlee: `Examples:` is only used by the bigframes team and it's their last content in the summary.

PR #361, opened 2024-03-27
feat: add summary page for Classes and Modules

dandhlee:
Curate summary pages given the entirety of the library content. Starting with Classes / Modules which are similar, then will expand to other types (Methods & Functions, and Properties & Attributes) in a followup.

Goes through all of the content, then extracts the necessary information, creating a separate `summary_class.yml` file. Adding as a separate top level entry. See live example: https://cloud.google.com/python/docs/reference/bigframes/latest/summary_class

Confirmed locally that the same page still gets produced, which was based off of #298, this PR just makes the code a bit more production quality.

`summary_overview` page is referenced but will have templates that live in the client libraries. See https://docs.google.com/document/d/1FmAOv9ald2W2Set8jPzN-gwQoE992LCWSu40KBVON28/edit?resourcekey=0-JLhux0549oJustMl46l3zA&tab=t.0 for design.

Test cases would be added for goldens, however it is currently disabled.

Towards b/263399076

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -28,7 +28,7 @@ import black
 import logging
 
 from collections import defaultdict
-from collections.abc import MutableSet
+from collections.abc import MutableSet, Mapping, Sequence
 from pathlib import Path
 from functools import partial
 from itertools import zip_longest
@@ -132,6 +132,21 @@ NOTICES = {
     BETA: 'Beta',
     PREVIEW: 'Preview',
     DEPRECATED: 'deprecated',
+}
+
+_SUMMARY_TYPE_BY_ITEM_TYPE = {
+    # Modules and Classes are similar types.
+    MODULE: CLASS,
+    CLASS: CLASS,
+}
+# Construct a mapping of name and content for each unique summary type entry.
+_ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE = {
+    summary_type: [[], []]  # entry name then entry content
+    for summary_type in set(_SUMMARY_TYPE_BY_ITEM_TYPE.values())
+}
+# Mapping for each summary page entry's file name and entry name.
+_FILE_NAME_AND_ENTRY_NAME_BY_SUMMARY_TYPE = {
+    CLASS: ('summary_class.yml', "Classes"),
 }
 
 # Disable blib2to3 output that clutters debugging log.
@@ -1348,6 +1363,57 @@ def pretty_package_name(package_group):
     return " ".join(capitalized_name)
 
 
+def _find_summary_details(
+    yaml_data: dict[str, Any],
+    summary_type: str,
+    cgc_url: str,
+) -> dict[str, Any]:
+    """Finds the summary details to add for a given entry."""
+    uid = yaml_data.get("uid", "")
+    item_to_add = uid if summary_type == CLASS else f"{uid}-summary"
+
+    if item_to_add not in _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE[summary_type][0]:
+        _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE[summary_type][0].append(
+            item_to_add
+        )
+
+    if summary_type in [CLASS]:
+        name_to_use = f"[{uid}]({cgc_url}{uid})"
+
+    return {
+        "uid": uid,
+        "name": name_to_use,
+        "fullName": uid,
+        "isExternal": False,
+    }
+
+
+def _render_summary_content(
+    children_name_and_summary_content: Sequence[Sequence[str]],
+    entry_name: str,
+    summary_type: str,
+    library_name: str,
+) -> Mapping[str, Any]:
+    """Returns the summary content in appropriate YAML format to write."""
+    summary_content = {}
+
+    if summary_type in [CLASS]:
+        summary_content = {
+            "items": [{
+                'uid': f'{summary_type.lower()}-summary',
+                'name': entry_name,
+                'fullName': f'{entry_name} Summary',
+                'langs': ['python'],
+                'type': 'package',
+                'summary': f'Summary of entries of {entry_name} for {library_name}.',
+                'children': children_name_and_summary_content[0],
+            }],
+            'references': children_name_and_summary_content[1],
+        }
+
+    return summary_content
+
+
 def find_uid_to_convert(
     current_word: str,
     words: List[str],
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -134,6 +134,21 @@ class Bcolors:
     DEPRECATED: 'deprecated',
 }
 
+_SUMMARY_TYPE_BY_ITEM_TYPE = {
+    # Modules and Classes are similar types.
+    MODULE: CLASS,
+    CLASS: CLASS,
+}
+# Construct a mapping of name and content for each unique summary type entry.
+_ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE = {
```
dansaadati, line 143: Can this be converted to a `dataclass`? Something like:
```
class SummaryEntry:
  name: str
  summary: str
```

For now it seems like there's just one summary type `CLASS`, so I think this information can be captured in just one data class. This would allow us to not use i-indexing below, which I think makes this more readable.

docfx_yaml/extension.py
```diff
@@ -134,6 +134,21 @@ class Bcolors:
     DEPRECATED: 'deprecated',
 }
 
+_SUMMARY_TYPE_BY_ITEM_TYPE = {
```
dansaadati, line 137: Nit: Maybe use `_SUMMARY_GROUP` instead? The phrasing can get a little bit confusing because you're working with Python types as well.

docfx_yaml/extension.py
```diff
@@ -1348,6 +1363,61 @@ def pretty_package_name(package_group):
...
+def _find_summary_details(
+    yaml_data: _yaml_type_alias,
+    summary_type: str,
+    cgc_url: str,
+) -> _yaml_type_alias:
+    """Finds the summary details to add for a given entry."""
+    uid = yaml_data.get("uid", "")
+    item_to_add = uid if summary_type == CLASS else f"{uid}-summary"
+
+    if item_to_add not in _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE[summary_type][0]:
```
dansaadati, line 1379: I think you can avoid this check if your use a `Set` for your `uid`s.

docfx_yaml/extension.py
```diff
@@ -1951,6 +2033,41 @@ def convert_module_to_package_if_needed(obj):
 
         file_name_set.add(filename)
 
+        for entry in yaml_data:
+            summary_type = _SUMMARY_TYPE_BY_ITEM_TYPE.get(entry.get("type"))
+            if not (summary_type):
+              continue
+
+            _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE[summary_type][1].append(
+                _find_summary_details(entry, summary_type, cgc_url)
```
dansaadati, line 2042: Nit:  I think because the return value of ` _find_summary_details` is only used to update the global mapping, I would just append the summary detail dict within the function.

docfx_yaml/extension.py
```diff
@@ -1897,6 +1964,21 @@ def convert_module_to_package_if_needed(obj):
...
+    pkg_toc_yaml.insert(
+        1,
+        {
+            "name": f"{app.config.project} APIs",
+            "items": [
+                {"name": "Classes", "href": "summary_class.yml"},
+            ],
+        }
+    )
+    cgc_url = (
```
dansaadati, line 1977: Nit: Move this below writing the toc file.

docfx_yaml/extension.py
```diff
@@ -1348,6 +1363,61 @@ def pretty_package_name(package_group):
...
+    if summary_type in [CLASS]:
+        summary_content = {
+            "items": [{
+                'uid': f'{summary_type.lower()}-summary',
+                'name': entry_name,
+                'fullName': f'{entry_name} Summary',
+                'langs': ['python'],
+                'type': 'package',
+                'summary': f'Summary of entries of {entry_name} for {library_name}.',
+                'children': children_name_and_summary_content[0],
```
dansaadati, line 1413: Nit: consider sorting by class/module name here.

docfx_yaml/extension.py
```diff
@@ -1348,6 +1363,61 @@ def pretty_package_name(package_group):
...
+    return {
+        "uid": uid,
+        "name": name_to_use,
+        "fullName": uid,
+        "isExternal": False,
+    }
+
+
+def _render_summary_content(
+    children_name_and_summary_content: Sequence[Sequence[str]],
```
dansaadati, line 1396: Could be mistaken here, but isn't this type `Sequence[Sequence[ str | dict ]` or `Sequence[Sequence[ str | _yaml_type_alias ]` ? The first sequence contains the list of uids, the second sequence contains the dict of YAML fields for references.

docfx_yaml/extension.py
dansaadati, line 1413: Doesn't this need to be converted into a YAML list?
Something like:
```
reduce(lambda x, y:  y + '\n - '  + x , children_name_and_summary_content[0])
```

dansaadati (changes requested):

docfx_yaml/extension.py
dandhlee: I think the problem was that I ran into the same items twice, not that the UIDs are not unique :/ I was seeing duplicate entries for each entry, but the UIDs are definitely unique (google.cloud.package.module.item)

docfx_yaml/extension.py
dandhlee: It's a bit tricky, I need to keep track of all the entries that's been submitted already. Either we need to keep an extra list, otherwise converting this to a dataclass we'd lose being able to have a sequence of information, instead we'll have `Sequence[SummaryEntry]` and not be able to easily go through the info :/

docfx_yaml/extension.py
dandhlee: This was a preferred style for mappings (summary_by_item[item] gives summary) - I could also just omit the types if that makes it better

docfx_yaml/extension.py
dandhlee: Nope, that gets handled already!

docfx_yaml/extension.py

docfx_yaml/extension.py
dandhlee: Done! Good point.

dandhlee (commented):
Verified that the changes did not affect produced result. Please take a look again!

dansaadati (approved):

PR #363, opened 2024-03-29
feat: add summary page support for methods and properties

dandhlee:
Continuation from #361. Adds support for Methods & Functions, Properties & Attributes.

Verified that the `summary_class.yml` is unaffected from this PR.

Tested and verified the result looks good, for staging in https://cloud.google.com/python/docs/reference/bigframes/latest/summary_class, https://cloud.google.com/python/docs/reference/bigframes/latest/summary_method and https://cloud.google.com/python/docs/reference/bigframes/latest/summary_properties.

Test cases would be added for goldens, however it is currently disabled.

Towards b/263399076.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -138,6 +138,12 @@ _SUMMARY_TYPE_BY_ITEM_TYPE = {
     # Modules and Classes are similar types.
     MODULE: CLASS,
     CLASS: CLASS,
+    # Methods and Functions are similar types.
+    METHOD: METHOD,
+    FUNCTION: METHOD,
+    # Properties and Attributes are similar types.
+    PROPERTY: PROPERTY,
+    ATTRIBUTE: PROPERTY,
 }
 # Construct a mapping of name and content for each unique summary type entry.
 _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE = {
@@ -147,6 +153,8 @@ _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE = {
 # Mapping for each summary page entry's file name and entry name.
 _FILE_NAME_AND_ENTRY_NAME_BY_SUMMARY_TYPE = {
     CLASS: ('summary_class.yml', "Classes"),
+    METHOD: ('summary_method.yml', "Methods"),
+    PROPERTY: ('summary_property.yml', "Properties and Attributes"),
 }
 
 # Disable blib2to3 output that clutters debugging log.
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -1384,12 +1392,51 @@ def _find_and_add_summary_details(
...
-        "uid": uid,
+        _ENTRY_NAME_AND_ENTRY_CONTENT_BY_SUMMARY_TYPE[summary_type][1].append({
+            "uid": uid,
+            "name": name_to_use,
+            "fullName": uid,
+            "isExternal": False,
+        })
+        return
+
+    # if summary_type in [METHOD, PROPERTY]:
```
dansaadati, line 1403: You can make this explicit with an `elif` statement, also you wouldn't need to return early above.

docfx_yaml/extension.py
```diff
@@ -1384,12 +1392,51 @@ def _find_and_add_summary_details(
...
+    short_name = yaml_data.get("name", "")
+    name_to_use = uid
+    if not (class_name := yaml_data.get("class", "")):
+        class_name = yaml_data.get("module", "")
+    anchor_name = f"#{'_'.join(class_name.split('.'))}_{short_name}"
```
dansaadati, line 1408: Does `class_name.replace('.', '_')` work here?

docfx_yaml/extension.py
```diff
@@ -1384,12 +1392,51 @@ def _find_and_add_summary_details(
...
+        summary.split(". ")[0],
+        summary.split(".\n")[0],
+        summary.split("\n\n")[0],
+        key=len,
+    )
+    if first_summary_line and first_summary_line[-1] != ".":
+      first_summary_line += "."
+
+    summary_to_use = (
+        f"{first_summary_line}\n\n"
```
dansaadati, line 1421: For the case where a summary isn't found, would this create extra whitespaces? I think this would become:

```
\n\n

See more: ...
```

dansaadati (approved):

docfx_yaml/extension.py
dandhlee: Chatted offline, keeping as-is.

docfx_yaml/extension.py
dandhlee: For methods and functions, it definitely helps to have the newlines with or without content after the syntax.
For properties & attributes, without the extra space is slightly more preferred but not a big deal. Also c.g.c will likely strip the newlines if its just blank content, it seems.

docfx_yaml/extension.py
dandhlee: 🤦  yes! Done.

dandhlee (commented):
Thank you!

PR #364, opened 2024-03-30
feat: grab repository metadata

dandhlee:
I've noticed an issue with the rest of the libraries not being able to properly find the library names from the given Sphinx info.

Seems like I need to extract the `.repo-metadata.json` file directly to source this information - Sphinx has very limited ways to deliver this information through `docs/conf.py`.

Based on the `googleapis/` [repo search](https://github.com/search?q=path%3A**%2F.repo-metadata.json+org%3Agoogleapis+%22%5C%22language%5C%22%3A+%5C%22python%5C%22%2C%22+%22%5C%22name%5C%22%3A+%22+%22%5C%22api_shortname%5C%22%3A+%22&type=code&ref=advsearch&p=1), it seems that using `name` suffices for shortname for the library. If I encounter ones that don't work, I'll create an exception around those few. I've sampled ~100 and seems to work with using `name` instead of `api_shortname`.

Verified that existing library generation demo for bigframes and logging still works.

If the repository metadata could not be retrieved, skip over getting the summary page details.

Fixes b/331977022.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -26,6 +26,7 @@ import copy
 import shutil
 import black
 import logging
+import json
 
 from collections import defaultdict
 from collections.abc import MutableSet, Mapping, Sequence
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -2086,9 +2106,11 @@ def convert_module_to_package_if_needed(obj):
         file_name_set.add(filename)
 
         for entry in yaml_data:
+            if not app.env.library_shortname:
+                continue
```
dansaadati, line 2110: This would apply to the whole library, right? If so, I think you can just `break` instead of `continue`.

docfx_yaml/extension.py
```diff
@@ -26,6 +26,7 @@
 import shutil
 import black
 import logging
+import json
```
dansaadati, line 29: Nit: import ordering

dansaadati (approved):

docfx_yaml/extension.py
dandhlee: Done. I was able to sort through the first party libraries.

docfx_yaml/extension.py
dandhlee: You're right, done.

dandhlee (commented):
Thank you! Verified that new changes did not change the result.

PR #374, opened 2024-05-02
feat: add support for enums

dandhlee:
Enums section was unformatted until now, adding a formatter for it.

I've considered stripping individual parts (name, value and description) to put into the table format, however we don't have any prior history of formatting enums this way. I've decided to just format the entire block in a codeblock, which is a significant improvement over what we have.

See b/195689267 for the staged example.

Mixed in a bit of lint/formatting fixes.

Fixes b/195689267.

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -400,6 +400,7 @@ def _parse_docstring_summary(summary):
     summary_parts = []
     attributes = []
     attribute_type_token = ":type:"
+    enum_type_token = "Values:"
     keyword = name = description = var_type = ""
 
     notice_open_tag = '<aside class="{notice_tag}">\n<b>{notice_name}:</b>'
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -486,7 +487,29 @@ def _parse_docstring_summary(summary):
 
         # Parse keywords if found.
         # lstrip is added to parse code blocks that are not formatted well.
-        if part.lstrip('\n').startswith('..'):
+        if (potential_keyword := part.lstrip('\n')) and (
+            potential_keyword.startswith('..') or
+            potential_keyword.startswith(enum_type_token)
```
dansaadati, line 492: Can this just be included in the [`extract_keyword`](https://github.com/googleapis/sphinx-docfx-yaml/blob/e147b144129d449cab6de0a8ca13c2588a9c0730/docfx_yaml/extension.py#L370) logic?

dansaadati (approved):

docfx_yaml/extension.py
```diff
@@ -400,6 +400,7 @@ def _parse_docstring_summary(summary):
     summary_parts = []
     attributes = []
     attribute_type_token = ":type:"
+    enum_type_token = "Values:"
```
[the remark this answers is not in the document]
dandhlee: If it was just `values` docstring tokenized I would have kept it up there but since we're looking for the specific `Values:` string I've kept it here.

docfx_yaml/extension.py
dandhlee: I thought about it, but unfortunately the entire section didn't get tokenized:
```
Values:
  name(val):
    ...
```
Other parts that do get tokenized and/or has proper format appear in this format:
```
.. note::
   ...

.. example::
   ...
```

dandhlee (commented):
Thank you!

PR #376, opened 2024-05-03
fix: format enum data into table format

dandhlee:
Instead of a codeblocked format for enum, use table formatting.

Corresponding doc-pipeline support for enum table in Python: https://github.com/googleapis/doc-pipeline/pull/642

See b/338431108 for a staged example.
Fixes b/338431108.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -400,8 +400,8 @@ def _parse_enum_content(parts: Sequence[str]) -> Sequence[Mapping[str, str]]:
     """Parses the given content for enums.
 
     Args:
-        parts: The content to parse, given in the form of separated by
-            newlines and left-indented.
+        parts: The content to parse, given in the form of a sequence of str,
+            which have been split by newlines and are left-indented.
 
     Returns:
         Sequence of mapping of enum entries for name and description.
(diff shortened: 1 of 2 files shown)


docfx_yaml/extension.py
```diff
@@ -396,11 +396,86 @@ def indent_code_left(lines, tab_space):
...
+    Returns:
+        Sequence of mapping of enum entries for name and description.
+
+    Raises:
+        ValueError: If the `Values` enum docstring is malformed.
+    """
+    enum_content: MutableSequence[Mapping[str, str]] = []
+    enum_name = ""
+    enum_description = []
+    prev_tab_space = -1
```
dansaadati, line 415: Is this used? I don't see it getting reassigned.

docfx_yaml/extension.py
```diff
@@ -396,11 +396,86 @@ def indent_code_left(lines, tab_space):
     return "\n".join(parts)
 
 
-def _parse_docstring_summary(summary):
+def _parse_enum_content(parts: Sequence[str]) -> Sequence[Mapping[str, str]]:
+    """Parses the given content for enums.
+
+    Args:
+        parts: The content to parse, given in the form of separated by
```
dansaadati, line 403: `in the form of a sequence of str, which have been split by newlines and are left-indented.`

docfx_yaml/extension.py
```diff
@@ -396,11 +396,86 @@ def indent_code_left(lines, tab_space):
...
+            new_entry = {
+                "id": enum_name,
+                "description": " ".join(enum_description),
+            }
+            if all(new_entry.values()):
+                enum_content.append(new_entry)
+            enum_description = []
+            # Only collect the name, not the value.
+            enum_name = part.split(" ")[0]
+        else:
```
dansaadati, line 438: Nit: I would remove a bit of conditional nesting here by checking for this condition first. I think it would be `if enum_name and enum_description and current_tab_space != 0)`.

dansaadati (approved):

docfx_yaml/extension.py
dandhlee: Ah yeah it's not really helping with empty input... I think this is unneeded.

docfx_yaml/extension.py
dandhlee: Good idea. Done.

dandhlee (commented):
Thank you!

PR #389, opened 2024-08-09
fix: retrieve keyword arguments for docstrings

### LINKED ISSUE #313: Optional parameters docstring not retrieved
When the method signature includes pseudo parameter to enforce positional/keyword arguments, those optional parameters are not included in the documentation.

Example:

```
tune_model(
    training_data: typing.Union[str, pandas.core.frame.DataFrame],
    *,
    train_steps: typing.Optional[int] = None,
    learning_rate: typing.Optional[float] = None,
    learning_rate_multiplier: typing.Optional[float] = None,
    tuning_job_location: typing.Optional[str] = None,
    tuned_model_location: typing.Optional[str] = None,
    model_display_name: typing.Optional[str] = None,
    tuning_evaluation_spec: typing.Optional[
        vertexai.language_models._language_models.TuningEvaluationSpec
    ] = None,
    default_context: typing.Optional[str] = None
) -> vertexai.language_models._language_models._LanguageModelTuningJob
```
only returns docstring for `traning_data`.

dandhlee:
From `inspect.getfullargspec`, keyword arguments were stored as different attributes, under `kwonlyargs` and `kwonlydefaults`.

* Added retrieval for `kwonlyargs`
* Retrieve for `kwonlydefaults` as well, for default values stored for keyword arguments. Interestingly, `defaults` are stored as tuples of values but `kwonlydefaults` is stored as name & value KVP, so I've directly addressed this. Since dictionaries retain order, this won't mess up the order for existing defaults algorithm.
* Adjusted spacing to remove indentation as much as possible for relevant parts 

Fixes #313. See b/358448617 for context and staging to verify the fixes.

- [x] Tests pass - this would be updated in goldens once it's fixed.

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -991,6 +991,8 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
                     # Find the first index which default arguments start at.
                     # Every argument after this offset_count all have default values.
                     offset_count = len(argspec.defaults)
+                    if argspec.kwonlydefaults:
+                        offset_count += len(argspec.kwonlydefaults)
                     # Find the index of the current default value argument
                     index = len(args) + count - offset_count
 
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -939,56 +939,75 @@ def _create_datam(app, cls, module, name, _type, obj, lines=None):
         if _type in [METHOD, FUNCTION, CLASS]:
             argspec = inspect.getfullargspec(obj) # noqa
             type_map = {}
-            if argspec.annotations:
-                for annotation in argspec.annotations:
-                    if annotation == "return":
-                        continue
-                    try:
-                        type_map[annotation] = _extract_type_name(
-                            argspec.annotations[annotation])
-                    except AttributeError:
-                        print(f"Could not parse argument information for {annotation}.")
-                        continue
 
-            # Add up the number of arguments. `argspec.args` contains a list of
-            # all the arguments from the function.
-            arg_count += len(argspec.args)
-            for arg in argspec.args:
-                arg_map = {}
-                # Ignore adding in entry for "self"
-                if arg != 'cls':
-                    arg_map['id'] = arg
-                    if arg in type_map:
-                        arg_map['var_type'] = type_map[arg]
-                        args.append(arg_map)
+            annotations = getattr(argspec, 'annotations', [])
+            for annotation in argspec.annotations:
+                if annotation == "return":
+                    continue
+                try:
+                    type_map[annotation] = _extract_type_name(
+                        argspec.annotations[annotation])
+                except AttributeError:
+                    print(f"Could not parse argument information for {annotation}.")
+                    continue
+
+            # Retrieve arguments from various source,s like `argspec.args`,
+            # `argspec.kwonlyargs` for positional/keyword arguments.
+            args_to_iterate = argspec.args
+            # Stores default information for kwonly arguments. Unlike
+            # inspect.defaults, which is a tuple of default values, kw defaults
+            # is a dict[name, defaultvalue].
+            kw_defaults = {}
+            if argspec.kwonlyargs:
+                args_to_iterate.extend(argspec.kwonlyargs)
+                if argspec.kwonlydefaults:
+                    kw_defaults.update(argspec.kwonlydefaults)
+            arg_count += len(args_to_iterate)
+            for arg in args_to_iterate:
+                # Ignore adding in entry for "self" or if docstring cannot be
+                # formed for current arg, such as docstring is missing or type
+                # annotation is not given.
+                if arg == 'cls' or arg not in type_map:
+                    continue
+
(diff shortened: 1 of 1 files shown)


docfx_yaml/extension.py
```diff
@@ -939,56 +939,77 @@ def _update_friendly_package_name(path):
...
+                    continue
+                try:
+                    type_map[annotation] = _extract_type_name(
+                        argspec.annotations[annotation])
+                except AttributeError:
+                    print(f"Could not parse argument information for {annotation}.")
+                    continue
+
+            # Retrieve arguments from various source,s like `argspec.args`,
+            # `argspec.kwonlyargs` for positional/keyword arguments.
```
tbpg, line 955:
```suggestion
            # Retrieve arguments from various sources like `argspec.args` and
            # `argspec.kwonlyargs` for positional/keyword arguments.
```

docfx_yaml/extension.py
```diff
@@ -939,56 +939,77 @@ def _update_friendly_package_name(path):
...
+            args_to_iterate = argspec.args
```
tbpg, line 956: Consider `original_args`.

docfx_yaml/extension.py
```diff
@@ -939,56 +939,77 @@ def _update_friendly_package_name(path):
...
-            for arg in argspec.args:
-                arg_map = {}
-                # Ignore adding in entry for "self"
-                if arg != 'cls':
-                    arg_map['id'] = arg
-                    if arg in type_map:
-                        arg_map['var_type'] = type_map[arg]
-                        args.append(arg_map)
+
+            annotations = getattr(argspec, 'annotations', [])
```
tbpg, line 943: Does this need tests?

docfx_yaml/extension.py
dandhlee: Done!

docfx_yaml/extension.py
dandhlee: Normally the test would involve updating the goldens as this function is called directly from Sphinx and can't be mocked easily.

I could refactor the logic out and add unit tests separately as a followup, and release only when the refactoring of this method is done.

dandhlee (commented):

docfx_yaml/extension.py
```diff
@@ -943,56 +943,77 @@ def _update_friendly_package_name(path):
...
+            # `argspec.kwonlyargs` for positional/keyword arguments.
+            original_args = argspec.args
+            # Stores default information for kwonly arguments. Unlike
+            # inspect.defaults, which is a tuple of default values, kw defaults
+            # is a dict[name, defaultvalue].
+            kw_defaults = {}
+            if argspec.kwonlyargs:
+                original_args.extend(argspec.kwonlyargs)
+                if argspec.kwonlydefaults:
+                    kw_defaults.update(argspec.kwonlydefaults)
```
dansaadati, line 968: Every entry in `argspec.kwonlydefaults` is contained in `argspec.kwonlyargs`, correct? The `argspec.kwonlydefaults` dict just gives us the default value? (just wanna check 😌)

docfx_yaml/extension.py
```diff
@@ -943,56 +943,77 @@ def _update_friendly_package_name(path):
...
+            arg_count += len(original_args)
+            for arg in original_args:
+                # Ignore adding in entry for "self" or if docstring cannot be
+                # formed for current arg, such as docstring is missing or type
```
dansaadati, line 972: For my understanding, how come a type  is necessary? Can the table (like on the cgc page) just be populated with the arg name to at least indicate it exists? If this is to drop the `*` from args, I think having that be skipped instead of *all* args with no type would be prefered.

docfx_yaml/extension.py
```diff
@@ -943,56 +943,77 @@ def _update_friendly_package_name(path):
...
-                        default_string = str(default)
-                        if 'at 0x' not in default_string:
-                            args[index]['defaultValue'] = default_string
-                # If we cannot find the argument, it is missing a type and was taken out intentionally.
-                except IndexError:
-                    pass
+            try:
+                defaults = getattr(argspec, 'defaults', [])
+                for count, default in enumerate(defaults):
+                    # Find the first index which default arguments start at.
```
dansaadati, line 995: Double checking this: line#974-975 wont throw off this offset, right? Like, could there be a default arg that get skips because it has no type?

dandhlee: Actually thanks for catching this - this should only apply for ones that appear before the keyword only arguments, and you are right, skipping them would have messed with the order, and worse - we don't actually know which one is skipped since the default values aren't paired with the parameters.

Instead, I've updated the code to (1) account for every arg into `args`, then (2) carefully place all default values for main args, then (3) remove any ones afterwards without any type.

dandhlee:
Ran tests over several libraries locally.

dansaadati (approved):
If possible, please make sure this has golden coverage somewhere (the `*` arg not breaking default args that follow).

PR #390, opened 2024-08-09
fix: ignore disambiguation for repeated entries

dandhlee:
When there is a duplicate but repeated entries in a row, it can cause confusion as there is no good disambiguator to return.

For example, given
```
google.cloud.run_v2.services
google.cloud.run_v2.services.services
```
this is not really possible to disambiguate - and we don't need to. The TOC / layout should make it clear which one is which.

Fixes b/296919376. See staged example there.

- [x] Tests pass

### DIFF
--- a/docfx_yaml/extension.py
+++ b/docfx_yaml/extension.py
@@ -162,6 +162,10 @@ _FILE_NAME_AND_ENTRY_NAME_BY_SUMMARY_TYPE = {
 logging.getLogger("blib2to3").setLevel(logging.ERROR)
 
 
+# Type alias used for yaml entries.
+_yaml_type_alias = dict[str, any]
+
+
 def _grab_repo_metadata() -> Mapping[str, str] | None:
     """Retrieves the repository's metadata info if found."""
     try:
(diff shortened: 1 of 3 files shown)


docfx_yaml/extension.py
```diff
@@ -1418,13 +1430,21 @@ def disambiguate_toc_name(pkg_toc_yaml):
...
         for part in module['uidname'].split("."):
-            if part not in name_entries[module_name]:
+            if part not in name_entries[module_name] or (
+                # If the name appears in a row, for example:
+                #   google.cloud.run_v2.services
+                #   google.cloud.run_v2.services.services
+                # then ignore this entry as this is clear from the left-nav layout.
+                part == prev_part
+            ):
                 name_entries[module_name][part] = 1
```
dansaadati, line 1444: I think this will work exclusively when the duplicate identifier is at the end. Like in your example, for `google.cloud.run_v2.services.services` it will see that `prev_part = 'services'` and reset the entry count to 1. However, if it was like `google.cloud.run_v2.services.services.jobs` then the entry count would go up to 2, bringing  it back into the TOC (if I'm not mistaken). Is that case a concern?

dansaadati (approved):

docfx_yaml/extension.py
dandhlee: The mapping would be like this:
```
'run_v2' = 1
'services' = 1
'jobs' = 1
```
in order for this to be a duplicate, we'd need another UID that ends with `jobs`. For example:

```
google.cloud.run_v2.services.services.jobs
google.cloud.run_v2.types.jobs
```
in this case, we can disambiguate with `types.jobs` and `services.jobs`.

Another example could be:

```
google.cloud.run_v1.services.services.jobs
google.cloud.run_v2.services.jobs
```
In this case we'd have `run_v1.jobs` and `run_v2.jobs` as disambiguators. `services` might also get disambiguated with `run_v1` or v2 prefixes, but in all cases this works. I _don't_ think I've left any other cases out. Sorry that this is confusing! :S

dansaadati: Got it -- in that case LGTM

PR #393, opened 2024-08-12
chore: use python3.10 for tests

vchudnov-g:


vchudnov-g:
https://github.com/googleapis/sphinx-docfx-yaml/pull/294 does not appear to have been enough.

tox.ini
```diff
@@ -12,10 +12,10 @@ commands =
 deps =
     {[testenv]deps}
 commands =
-    python3 -m pytest tests
+    python3.10 -m pytest tests
```
dandhlee, line 15: Instead of specifying them here, we can globally set them like this 

```
[tox]
envlist = 
    tests
    py3.10
```

dandhlee:
Ah, the issue with tests not running is because GH actions is not automatically triggering them, not because of versions issue.

vchudnov-g:
> Ah, the issue with tests not running is because GH actions is not automatically triggering them, not because of versions issue.

So we should undraft https://github.com/googleapis/sphinx-docfx-yaml/pull/340 then, right?

dandhlee:
Yes, just the `ci.yaml` file though - and I remember we had updated guidelines on being able to invoke that recently.

dandhlee (approved):
Fine to update this file for now. Thank you!

PR #410, opened 2025-03-05
chore: remove unused files

parthea:
See https://github.com/googleapis/synthtool/pull/2067 and https://github.com/googleapis/synthtool/pull/2035

Note: It's not clear how [.kokoro/docs/generate-docs.cfg](https://github.com/googleapis/sphinx-docfx-yaml/blob/main/.kokoro/docs/generate-docs.cfg) is used and if it should be deleted


dandhlee (approved):

dandhlee:
`.kokoro/docs/generate-docs.cfg` is used for one-off jobs for all doc generations for Cloud libraries, which uses https://github.com/googleapis/sphinx-docfx-yaml/blob/main/.kokoro/generate-docs.sh.

